PR #2, opened 2013-10-09
fixed typo in iscsi_discovery usage()

MagicalTux:
Noticed that the help for iscsi_discovery had a typo

changing "trasnpot" into "transport" so it looks better


gonzoleeman:
Doesn't seem to rise to the level of a fix that's actually worth doing.

cleech:
Thank you, this was rolled up into a larger merge by the Debian package maintainer who was carrying it

PR #5, opened 2014-05-14
Update mgmt_ipc.c

jduanen:
Ran into a problem where iscsiadm was closing an already closed fd (returning EBADF).
Seems like the close() in line 466 is redundant as it is done in mgmt_ipc_destroy_queue_task().
Could also assign qtask->mgmt_ipc_fd to NULL, but it seems better to do it this way.


mikechristie:
Thanks for tracking this down and also providing a patch jduanen!

This was posted to the list and merged upstream via that proccess, so I am closing this request here.

PR #9, opened 2015-01-19
Kernel include path

xypron:
The path to the kernel include files is given by
/lib/modules/`uname -r`/build
on Debian.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/README
+++ b/README
@@ -92,7 +92,7 @@ by running:
 	make kernel
 
 When building those modules the kernel source found at
-/lib/modules/`uname -a`/build
+/lib/modules/`uname -r`/build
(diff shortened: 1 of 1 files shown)


agrover:
I don't think Mike uses github for patch review. Would you please send it to open-iscsi@googlegroups.com? Or I'd be happy to, if it would be easier.

mikechristie:
I guess it's time for me to get used to at least do trivial reviews/merges in github since we get a lot of doc fixups this way.

Thanks for the patch xypron, and thanks for watching out agrover.

PR #12, opened 2015-09-08
add systemd unit for automatic target login

eworm-de:


gonzoleeman:
I believe this is missing a few items. Please get agreement on the proper format for this file on the mailing list then resubmit, if needed.

PR #20, opened 2016-07-20
iscsid: reset socket when session recovery

glzhao:
Close without linger setting would not reset socket, corrupted data
could been sent to target server when tcp retry.

Signed-off-by: Guangliang Zhao <lucienchao@gmail.com>


gonzoleeman:
Chris, what's the status on this?

cleech:
This was fixed in a slightly different form that also checked connection state to only apply SO_LINGER when closing the socket outside of the normal iSCSI logout process.

PR #34, opened 2016-12-02
These are the "cleanup shutdown" changes discussed and approved on the mailing list

gonzoleeman:


gonzoleeman:
Any objections to merging this? I changed "event_loop_stop" to volatile, the only edit suggested. I did that as a separate commit.

PR #36, opened 2017-01-03
fix musl on open-iscsi

prometheanfire:
Portions of this patch have been taken from Alpine [1] and Gentoo [2].

This patch has been tested on glibc and musl systems.

[1] http://git.alpinelinux.org/cgit/aports/log/main/open-iscsi/musl-fixes.patch
[2] https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=da400dec15a2987751259860fea158bf252b3baa

### DIFF
--- a/iscsiuio/src/unix/libs/bnx2x.c
+++ b/iscsiuio/src/unix/libs/bnx2x.c
@@ -42,7 +42,7 @@
 #include <arpa/inet.h>
 #include <linux/types.h>	/* Needed for linux/ethtool.h on RHEL 5.x */
 #include <linux/sockios.h>
-#include <linux/ethtool.h>
+#include "ethtool-compat.h"
 #include <sys/mman.h>
 #include <sys/ioctl.h>
 #include <sys/types.h>
--- a/iscsiuio/src/unix/libs/ethtool-compat.h
+++ b/iscsiuio/src/unix/libs/ethtool-compat.h
@@ -1,0 +1,51 @@
+/*
+ * ethtool-compat.h: adopted from
+ # ethtool.h: Defines for Linux ethtool.
+ *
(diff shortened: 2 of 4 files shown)


gonzoleeman:
The changes to report an error if an unknown option is supplied seem unrelated.

blueness:
@gonzoleeman Yeah this pull request comes from two sources and addresses different issues.  I did the ethtool-compat.h part.

I don't get the alpine linux fix which is what you're questioning. They're claiming it addresses their bug: http://bugs.alpinelinux.org/issues/4802.  It appears to be a difference between how getopt_long() behaves in musl and glibc.  We should probably talk to musl upstream before merging this as it may screw up iscsiadm's behavior.

For what its worth, I did test my patch on glibc and its fine.

gonzoleeman:
It looks like musl sets "optopt" in the getopt_long(3) code now, and that breaks iscsiadm's assumptions. I would suggest making that a separate issue and resubmitting your changes alone for this one. Or do them one at a time. :)

prometheanfire:
ya, too bad github doesn't know how to stack patches for individual review.  I'll revert the changes to `usr/iscsiadm.c` and re-commit them once this merges.

prometheanfire:
I created https://github.com/open-iscsi/open-iscsi/pull/37 that's stacked on this

gonzoleeman:
Perhaps I did not make my preference request strong enough, so let me make it clear now. You will have to withdraw this request and resubmit with the issues raised above, i.e. (1) unrelated error reporting changes, and (2) the optopt changes. Apologies for not making this clear sooner.

Also, please note that you do not have to stack patch requests. You can use a different branch for each pull request. If there are no conflicts between the requests this works quite well.

prometheanfire:
closing in favor of #75

PR #42, opened 2017-03-02
Unify README headings.

m4z:
This commit…
- promotes section 1.1 "Features" to a full-fledged heading,
- removes unnecessary indentation in section 1.1,
- adds a uniform amount of newlines before (2 lines) and after each heading+"underline" (1 line),
- replaces all heading "underlines" with equals signs (some where minuses before),
- removes some trailing colons and whitespace from headings.


cleech:
Thanks for the README improvements

PR #45, opened 2017-03-02
Improve README punctuation.

m4z:
(And fix a few typos.)

### DIFF
--- a/README
+++ b/README
@@ -66,7 +66,7 @@ iscsid, and a management utility iscsiadm.
 ===============
 
 As of today, the Open-iSCSI Initiator requires a host running the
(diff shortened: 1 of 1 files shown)


cleech:
Looks like this now conflicts with your earlier changes, want to resolve it?

(for future reference you can have multiple change commits on one branch and send then in a single pull request when they're all related like this)

m4z:
Yeah, seems like this noob right here got lost in his own branches. Should be fixed now.

I still have some potentially controversial changes coming up though, so I deemed it better to split the different kinds of changes. Will see if I can keep the branches to a minimum. :wink:

PR #51, opened 2017-04-18
    C/Python API -- libiscsi

cathay4t:
C/Python API -- libiscsi
    
     * Just porting the patch used by RHEL/Fedora to upstream with needed changes.
     * Shipping internal share library `libiscsi_usr.so.0`, so that
       libiscsi.c don`t have to worry about what object to link against.
     * Test cases are stored in folder `tests` and could be kicked off by
       `make check` command.
     * No `make install` for libiscsi. Package maintainer should do so.
     * No install or compile command for libiscsi python module, package
       maintainer could use
       `python setup.py build && python setup.py install`


cathay4t:
Tested on tcp iface and a be2iscsi iBFT.

gonzoleeman:
Hello. Chris may already be familiar with this code, but I am not. Can you summarize what this does. A quick scan makes it look like (1) you are adding a C API (is python also in there?) for clients to use to manipulate iSCSI state instead of having to use iscsiadm, and (2) some tests that go with those new APIs. Is this correct? Is there documentation on the new APIs? How long have they been in use in your distribution (i.e. have they already been tested in the field)? Do the tests just validate the library APIs, or can they be used to test open-iscsi in general?

Unless Chris jumps in and says otherwise, I'd like to get a copy of your branch and look at the code further before approving.

cathay4t:
This PR just include initial patch for this C and Python API.
Please check `libiscsi/TODO` file for things will be fixed in coming patches.

Summery on this PR:
 * C API(libiscsi.h).
   Build: `make`
   Install: `install -pm 755 libiscsi.so* /usr/lib64/; installl libiscsi.h /usr/include/`
   Use: simply add `LDFLAGS += -liscsi` to makefile.
 * Python 2 and Python 3 module `libiscsi`. 
    Build: `python setup.py build`
    Install: `python setup.py install`
    Use: `python -c 'import libiscsi'`.

Implementation:
 * Using functions from `usr` folder which is used for `iscsiadm`, `iscsid`.
 * Created internal library `libiscsi_usr.so` to expose functions in `usr` folder.
 * The original code is not coded by myself, I just did some Makefile, rebase and code review.
    Since RHEL and Fedora are already used it for a quit a long time, I cannot change its ABI.

Documents:
 * `make html` will generate html document using doxygen.

Functions:
 * Use `SendTarget` to discovery iscsi nodes with/without discovery authentication.
 * Use PPC or iBFT firmware information to discovery iscsi nodes.
 * Change/set iscsi session authentication.
 * Login and logout iscsi node.
 * Query PPC or iBFT firmware IQN and network config.

Test status:
 * This patch was included in RHEL 7 and Fedora since 2012 and only been used for anaconda 
    ever since. Anaconda as an OS install tool does not care about query current iSCSI session,
    hence this PR only include `discovery then login` functions. Storaged(Now merged with udisks)
    also use it for their iscsi module, but it seems not really used/tested in field.
 * `make check` will run in code test, please check `libiscsi/tests/TEST_SETUP.txt` for setup.
    We could create a script to fit into CI.

cathay4t:
Q: is python also in there?
A: Yes. Both Python2 and Python3 are supported. iscsiadm, 
Q: Is this correct? 
A: The test cases are only for C library functions.
Q: Is there documentation on the new APIs? 
A: `make html` will do.
Q: How long have they been in use in your distribution (i.e. have they already been tested in the field)?
A: It was introduced into RHEL and Fedora 2012 by the date noted in original patch. It is only used by anaconda(redhat install tool), I will not call it as tested by field.
Q: Do the tests just validate the library APIs, or can they be used to test open-iscsi in general?
A: Just verify the library API. We need more work(like check return and handle expected error) to cover the general testing.

cleech:
I'm familiar with this code, in so much as I've managed to keep it working in Fedora and RHEL but haven't put effort into further expanding on it.  I know we had conversations about upstreaming it before, but Mike was of the opinion that it would be better to rewrite the core code as a library and make the open-iscsi tools use it.  As is, this exposes some of the internal code that wasn't necessarily written with safety as a library interface in mind.

I'm willing to revisit that, as there hasn't been progress on making a better library interface to this stuff, but we should do so on the mailing list instead of here I think.

cathay4t:
@cleech I start a email thread in mailing list for further discuss. https://groups.google.com/forum/#!topic/open-iscsi/KylFp8Ohyi4

carmark:
@cathay4t  Will this API follow the [IMA](https://www.snia.org/tech_activities/standards/curr_standards/ima)?

cathay4t:
@carmark No. This library is intend to expose all features of open-iscsi, not for vendor iscsi HBA management.

cathay4t:
@gonzoleeman @cleech It has been radio silence here and in mailing list for a while. Any suggestion what approach should we take for C and Python API of open-iscsi?

gonzoleeman:
> On May 22, 2017, at 9:59 PM, Gris Ge <notifications@github.com> wrote:
> 
> @gonzoleeman @cleech It has been radio silence here and in mailing list for a while. Any suggestion what approach should we take for C and Python API of open-iscsi?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

I must admit I’ve been thinking about this and somewhat agree with Mike’s original complaint, i.e. that these APIs will duplicate functionality already in the daemon or the admin command, and that the _right_ way to do this would be to rewrite parts of iscsid/iscsiadm as part of this effort.

In other words, we really need to work on the guts instead of pasting another layer on them.

But I’m not sure how strongly I feel about this. Chris?

— 
Lee Duncan

cathay4t:
@gonzoleeman You mean create a libiscsi_usr library which could be used by iscsiadm/iscsid and external library user?

gonzoleeman:
Sorry for my tardy response. Yes, that's exactly what I mean.

cathay4t:
@gonzoleeman Thanks. I will working on new PRC for this.

cathay4t:
Closing as I will create new PR with different approach to C/Python API of open-iscsi.

PR #55, opened 2017-06-02
Add bnx2{,x} software initiator IBFT support

apatters:
Add support in iscsistart for bnx2{,x} software initiators. This allows boot from iSCSI for devices that either cannot support iSCSI hardware offload or are configured in firmware not to support it.

The scsistart -N option will now bring up the interface (but not configure it) as specified in the IBFT for hardware offload devices.

Note: I cannot find a way to programmatically determine whether a bnx2{,x} device is configured for
hardware or software, so we check for the presence of the iscsiuio executable as an indication that hardware offload is being used.

### DIFF
--- a/include/iscsi_net_util.h
+++ b/include/iscsi_net_util.h
@@ -2,6 +2,7 @@
 #define __ISCSI_NET_UTIL_h__
 
 #define ISCSI_HWADDRESS_BUF_SIZE 18
+#define ISCSIUIO_PATH "/sbin/iscsiuio"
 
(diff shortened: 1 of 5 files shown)


gonzoleeman:
This looks fine to me, except I also dislike having to check for the static existence of a binary to detect support for hardware offload support. Shouldn't iscsiuio actually be running and not just present?

Also, a nit: the indentation seems a little funky in two of the patches, or is it just me? Lastly, you could have submitted the 3rd patch -- the man page fix -- separately, as it has nothing to do with the other two. But it's fine if you want to bundle it, too.

PR #57, opened 2017-06-09
Fixed typo for spelling of 'default'

y011:


gonzoleeman:
This is an acceptable first patch, but in general we try to address more serious issues, though misspellings are annoying.

y011:
Yeah I hear you. Figured I'd submit it while I was thinking about it. I was just using a package of open-iscsi 2.0.873-3ubuntu9 and noticed those and another typo that I saw was later corrected in 2016:

```# iscsi_discovery 
Usage: /usr/sbin/iscsi_discovery <IP> [-p <port>] [-d] [-t <tcp|iser> [-f]] [-m] [-l]
Options:
-p		set the port number (defualt is 3260).
-d		print debugging information
-t		set trasnpot (default is tcp).
-f		force specific transport -disable the fallback to tcp (default is fallback enabled).
			force the transport specified by the argument of the -t flag.
-m		manual startup - will set manual startup (default is automatic startup).
-l		login to the new discovered nodes (defualt is false).

PR #66, opened 2017-11-13
Rephrased "only valid for but not only"

poige:


gonzoleeman:
No, I don't think this is correct. You've left off the piece of information that works in Session mode but does nothing.

poige:
who needs this piece?

poige:
Given "it does nothing" it shouldn't clutter the docs (until it starts to do a thing)

gonzoleeman:
No, you are missing my point. The man page is trying to get across the fact that this option is *accepted* (i.e. no syntax error) by the iscsiadm command in session mode but has no functionality.

It certainly is not good grammer, but it's accurate. Perhaps a rephrasing like:

"This option is only valid and functional for node mode. The option is accepted but ignored for session mode, for historical reasons."

?

poige:
> ?

It's a garbage in the docs until functionality is there. — Awkwardly expressed one. No one needs it in essence. It's not a memo for developers, remember?

poige:
> not garbage, it is information

Information not needed is what they call "garbage". It's overloading man page w/o any real value in return.

gonzoleeman:
Thank you for your opinion. I will wait for Chris' opinion before closing this.

poige:
Yeah; you're giving the info users don't need. Developers might need this info but users' manual isn't proper place. man page should be as concise as possible — it's _for users_; not for people who develop it. From my experience (I read lots of man typically) `iscsiadm`'s man page can't be called any good example — overloaded, poorly organized and redundant (with copy-pasts like this). Each excessive part that can be removed from it, specially about "nop" like this, is better be removed if you want user to have better experience with it and the tool set consequently.

poige:
> If I run "iscsiadm -m session -l", and it does nothing, I am going to look in the man page.

go look; you'll find it's said it's only functional in another mode, what's wrong with you?

gonzoleeman:
You are a troll.

poige:
And you a hero. No surprises

gonzoleeman:
I still welcome a better re-write (I've created a new issue on the man page in general) or fixing the iscsiadm command to print a warning in this case. Both more useful than just removing information IMHO.

poige:
Doesn't make any sense to end users. Users do Not need this piece of info. That's the thing.

PR #72, opened 2017-12-15
Iscsiuio fixes

gonzoleeman:
This is a series of commits that fix vulnerability issues found by Qualsys in the iscsiuio subsystem, used to talk to broadcom offload cards. There _should_ be no functional changes for normal operations. The result was sanity tested on a system with such an adapter, but not extensively tested (as we still have no good test suite).


gonzoleeman:
@cleech I'd like to get your feedback on these changes, as I suspect you may have your own (different) set to address this?

PR #75, opened 2017-12-31
make open-iscsi compatable with musl-libc

prometheanfire:
Use argerror instead of optopt for musl compatibility
  It looks like musl sets "optopt" in the getopt_long(3) code now, and
  that breaks iscsiadm's assumptions.  So change to use argerror instead.

Include if_ether from the installed libc instead of linux_headers, this also
imports from linux_headers and helps ensure libc compatability.

Include fcntl.h as it's needed on musl.

Fix sort line to work on musl (__compar_fn_t is not available).

### DIFF
--- a/iscsiuio/src/unix/libs/bnx2x.c
+++ b/iscsiuio/src/unix/libs/bnx2x.c
@@ -36,6 +36,7 @@
  * bnx2x.c - bnx2x user space driver
  *
  */
+#include <netinet/if_ether.h>
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
--- a/usr/idbm.c
+++ b/usr/idbm.c
@@ -25,6 +25,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <dirent.h>
 #include <limits.h>
 #include <sys/stat.h>
--- a/usr/iscsi_net_util.c
+++ b/usr/iscsi_net_util.c
@@ -31,7 +31,7 @@
 #include <linux/sockios.h>
 #include <linux/if_vlan.h>
 #include <net/if_arp.h>
-#include <linux/if_ether.h>
+#include <netinet/if_ether.h>
 
 #include "sysdeps.h"
 #include "ethtool-copy.h"
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -3265,6 +3265,7 @@ main(int argc, char **argv)
 	int tpgt = PORTAL_GROUP_TAG_UNKNOWN, killiscsid=-1, do_show=0;
 	int packet_size=32, ping_count=1, ping_interval=0;
 	int do_discover = 0, sub_mode = -1;
+	int argerror = 0;
 	int portal_type = -1;
 	int timeout = ISCSID_REQ_TIMEOUT;
 	struct sigaction sa_old;
@@ -3431,6 +3432,10 @@ main(int argc, char **argv)
 			break;
 		case 'h':
 			usage(0);
+			break;
+		case '?':
+			log_error("unrecognized character '%c'", optopt);
+			argerror = 1;
 		}
 
 		if (name && value) {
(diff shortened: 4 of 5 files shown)


prometheanfire:
This was taken directly from alpine, let me know if you want it split (based on the paragraphs of the commit message).  https://git.alpinelinux.org/cgit/aports/tree/main/open-iscsi/musl-fixes.patch

gonzoleeman:
I have tried to build open-iscsi latest with musl on openSUSE Leap 42.2, and it failed miserably. The fatal problem seems to be that the include file <linux/types.h> is used in several places in open-iscsi but is not available in musl.

I hesitate to merge any musl changes if musl is not gong to ever work for open-iscsi.

Do you have your own branch some place that works with musl that I could test?

prometheanfire:
sys-kernel/linux-headers (/usr/include/linux/types.h)

At least on gentoo we have that as part of our system (on both musl and glibc).  And it is required for open-iscsi as well.

gonzoleeman:
Yes, of course Linux boxes have that include file in /usr/include/linux/types.h, but I have to add "-I/usr/include" to get it on the musl build, and that opens a can of worms on openSUSE that can't be closed.

This just does not compile using the latest upstream musl on SLES/openSUSE, and I suspect the same would be true of other Linux distros. So this pull request is a no-go.

You'll have to make this work on musl (linux or otherwise) and on linux/gcc for it to become part of the main development line.

richfelker:
I think there's a bit of a misunderstanding here. musl libc packages available in non-musl-based distros like openSUSE or Debian generally do not include a complete cross compiling environment with additional third-party libraries or headers (including the linux headers). If you want these things you have to install them yourself, but more to the point, these sorts of packages are best thought of as a simple introductory/evaluation install of musl. They are **not** what you want to be testing application compatibility against.

The actually-meaningful targets are musl-based distros like Alpine Linux or Void Linux. These of course have packages for the linux kernel headers and any other dependency libraries you might need. As I understand it, Alpine Linux is presently using and packaging open-iscsi just fine with a few patches they include for the distro package file, and they should be reasonable for upstream. If not, please indicate why.

gonzoleeman:
Okay, I get it a little better now, but I'm still not sure about this.

My first concern of course is that any changed added for musl would not interfere with the major (gcc) usage of the package.

Second, I'm concerned that I cannot easily check any changes suggested for musl. I just spent a couple of hours trying to download and test on Alpine linux. I admit it's a distro I know nothing about, but still it was not easy to figure out. I finally gave up on building an environment where I could compile and test open-iscsi, simply because I don't have the time. And I don't want to make the commitment to test on such a platform regularly, based on that experience. So then the issue is whether or not Chris and I are comfortable with changes we cannot test on the platform where they are targeted. Unless you can suggest an easier way to test changes on this platform ...

prometheanfire:
I have an openstack image that could help, or if that doesn't work I can build a more generic image that you could boot with libvirt/qemu.

http://distfiles.gentoo.org/experimental/amd64/openstack/gentoo-openstack-amd64-hardened-musl-latest.qcow2

gonzoleeman:
No, I think the problem is more fundamental than just being able to test that the changes compile. For example, I am working on the built-in open-iscsi test suite, to try to get it running again. That test suite requires some support programs that may not be available on musl. Who will take ownership of such issues? I just don't have the time to add this to my list. Also, as a separate issue you submitted shows, there are changes you'd like to make that are too pervasive, such as adding compatibility headers to replace Linux include files, which is an approach that adds too much maintenance work down the line (required to track changes in the real Linux include file).

For these reasons I believe supporting musl is something that can't easily be done right now. I fear that those that need open-iscsi on musl will create there own GitHub branch from open-iscsi so that they can easily track and merge new changes. I know from having to do that for SUSE that each distribution has some changes they must maintain that aren't appropriate upstream, so this is a task I assume the musl folks are already undertaking.

Perhaps you and I and Chris can discuss this issue and find a better long-term solution. If so, contact me off-list or use the open-iscsi mailing list. I have been thinking about trying to set up an open-iscsi mini-summit some time soon, so perhaps discussing this would be appropriate for that?

prometheanfire:
Ya, we already have open-iscsi working on gentoo and alpine.  For gentoo it was needed for openstack cinder.  If automated gating is implemented that'd help at least one of your objections (that it's not tested to your standards).

richfelker:
@gonzoleeman: If those are your concerns, don't frame it as "adding support for musl" and don't claim "support" at this time. The proposed patches are purely portability fixes. What's important is that you be able to test that they don't break anything on platforms you do explicitly support, and I think that's relatively easy.

PR #76, opened 2017-12-31
use compat <linux/ethtool.h> to build on musl

prometheanfire:
Needed to build on musl, tested on Gentoo.

### DIFF
--- a/iscsiuio/src/unix/libs/bnx2x.c
+++ b/iscsiuio/src/unix/libs/bnx2x.c
@@ -42,7 +42,7 @@
 #include <arpa/inet.h>
 #include <linux/types.h>	/* Needed for linux/ethtool.h on RHEL 5.x */
 #include <linux/sockios.h>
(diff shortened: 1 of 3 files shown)


gonzoleeman:
On hold until I have a better understanding of how to build open-iscsi with musl.

gonzoleeman:
musl support is on hold, so not sure if this pull request is worth it, since it can't solve the problem of musl compilation on linux not working, even with this fix.

So planning on also closing this pull request unless I hear otherwise.

gonzoleeman:
This is I believe the kind of change which I would not like to see in open-iscsi, since it replaces a perfectly good system include file with a local copy so that open-iscsi can compile on a platform where that perfectly-good include file is not present. That means a lifetime of tracking changes from the system include file in the local copy. Not a good thing.

gonzoleeman:
Sorry, as stated above this kind of change is not compatible with non-musl open-iscsi

PR #78, opened 2018-01-16
Ignore common build output files

gonzoleeman:
After years of see these build files listed when I do "git status", I'm ready to add them to the ignore list.


gonzoleeman:
Seems reasonable, as long as the .gitignore file doesn't get crazy.

PR #79, opened 2018-01-16
Automate logging into iSCSI FW targets.

gonzoleeman:
Add a script that logs into all iSCSI firmware
targets, and add a udev rule to call that script
when new targets are detected. Also, add a
man page for the script.


gonzoleeman:
This just adds a new util program which may be handy for udev usage.

PR #80, opened 2018-01-17
Repo cleanup, delete old kernel stuff and unused stub code

cleech:
Old stuff that's been unused forever; out of tree kernel patches from before the upstream acceptance around 2005, and unused stub code from a plan to support BSDs with the same tools that never happened.


gonzoleeman:
Thank you for doing this Chris!

PR #89, opened 2018-02-03
musl build fixes

kraj:


gonzoleeman:
I will compile test these, but they all look reasonable at first glance.

kraj:
@gonzoleeman any updates ?

gonzoleeman:
Please make just the one change I requested, and I'll be glad to merge these into master

kraj:
@gonzoleeman the reason is there in the commit message

```
including nic.h before linux/ethtool.h avoids redefinitions of
eth structs
```

gonzoleeman:
People editing the code don't normally look at the git commit comments that added the code. Since the local includes traditionally go below the global includes, I can see somebody (like me) moving the include back down below, especially since it wouldn't break any non-musl compilers. So please add a simple comment, something like your commit comment? Is that asking too much?

kraj:
Oh thats not a problem. It was not clear to me from earlier comment if you wanted the comment in the code itself. Let me do that and refresh the PR

kraj:
@gonzoleeman done, please review and let me know if it needs some more work

PR #92, opened 2018-02-24
iscsid: fix logging level when starting daemon

az0uz:
Changing log level for daemon start message.
This is not an error, just an info.


gonzoleeman:
Thank you!

PR #93, opened 2018-03-12
 libopeniscsiusr: Add full iscsi interface support

cathay4t:
* Improve existing iface query from sysfs.
    
     * Add new functions to query iSCSI interface information from
       /etc/iscsi/ifaces/ folder:
        * iscsi_ifaces_get()/iscsi_ifaces_free()
        * iscsi_iface_get()/iscsi_iface_free()
        * iscsi_iface_dump_config()
        * iscsi_iface_print_config()
        * iscsi_is_default_iface()
    
     * Use above functions in iscsiadm for iface query.
    
     * The `iscsiadm -m iface -P 1` command is untouched as it require node
       query.
    
     * Bump libopeniscsiusr library version to 0.1.1 due to API adding.
    
     * Tested on qla4xxx, cxgb4i, bnx2i, qedi and of course iscsi_tcp.

### DIFF
--- a/libopeniscsiusr/iface.c
+++ b/libopeniscsiusr/iface.c
@@ -36,6 +36,7 @@
 #ifdef USE_KMOD
 #include <libkmod.h>
 #endif
+#include <limits.h>
 
 #include "libopeniscsiusr/libopeniscsiusr.h"
 #include "misc.h"
--- a/iscsiuio/src/unix/libs/bnx2x.c
+++ b/iscsiuio/src/unix/libs/bnx2x.c
@@ -36,6 +36,11 @@
  * bnx2x.c - bnx2x user space driver
  *
  */
+
+/* include nic.h before linux/ethtool.h to avoid redefinitions of
+ * eth structs
+*/
+#include "nic.h"
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
@@ -58,7 +63,6 @@
 #include "bnx2x.h"
 #include "cnic.h"
 #include "logger.h"
-#include "nic.h"
 #include "nic_id.h"
 #include "nic_utils.h"
 #include "options.h"
--- a/iscsiuio/src/unix/libs/qedi.c
+++ b/iscsiuio/src/unix/libs/qedi.c
@@ -49,7 +49,6 @@
 #include <arpa/inet.h>
 #include <linux/types.h>
 #include <linux/sockios.h>
-#include <linux/ethtool.h>
 #include <linux/netlink.h>
 #include <sys/mman.h>
 #include <sys/ioctl.h>
--- a/iscsiuio/src/unix/nic_utils.c
+++ b/iscsiuio/src/unix/nic_utils.c
@@ -305,7 +305,7 @@ static int nic_util_enable_disable_multicast(nic_t *nic, uint32_t cmd)
 	/* Prepare the request */
 	memset(&ifr, 0, sizeof(ifr));
 	strncpy(ifr.ifr_name, nic->eth_device_name,
-		sizeof(nic->eth_device_name));
+		sizeof(ifr.ifr_name));
 	memcpy(ifr.ifr_hwaddr.sa_data, multicast_addr.addr, ETH_ALEN);
 
 	fd = socket(AF_INET, SOCK_DGRAM, 0);
--- a/libopeniscsiusr/Makefile
+++ b/libopeniscsiusr/Makefile
@@ -49,7 +49,7 @@ LIBADD =
 all: $(LIBS) $(LIBS_MAJOR) $(TESTS) doc
 
 $(LIBS): $(OBJS)
-	$(CC) $(LDFLAGS) -shared -Wl,-soname=$@ -o $@ $(OBJS) $(LIBADD)
+	$(CC) $(CFLAGS) $(LDFLAGS) -shared -Wl,-soname=$@ -o $@ $(OBJS) $(LIBADD)
 	ln -sf $@ $(DEVLIB)
 
 $(LIBS_MAJOR): $(LIBS)
--- a/libopeniscsiusr/sysfs.c
+++ b/libopeniscsiusr/sysfs.c
@@ -325,6 +325,8 @@ int _iscsi_host_id_of_session(struct iscsi_context *ctx, uint32_t sid,
 	struct dirent **namelist = NULL;
 	int n = 0;
 	const char *host_id_str = NULL;
+	const char iscsi_host_dir_str[] = "/iscsi_host/";
+	const unsigned int iscsi_host_dir_strlen = strlen(iscsi_host_dir_str);
 
 	assert(ctx != NULL);
 	assert(sid != 0);
@@ -339,8 +341,16 @@ int _iscsi_host_id_of_session(struct iscsi_context *ctx, uint32_t sid,
 				 _SYSFS_DEV_CLASS_ISCSI_SESSION, sys_dev_path),
 	      rc, out);
 
-	snprintf(sys_scsi_host_dir_path, PATH_MAX, "%s/iscsi_host/",
-		 sys_dev_path);
+	if ((strlen(sys_dev_path) + iscsi_host_dir_strlen) >= PATH_MAX) {
+		rc = LIBISCSI_ERR_SYSFS_LOOKUP;
+		_error(ctx, "Pathname too long: %s%s",
+		       sys_dev_path, iscsi_host_dir_str);
+		goto out;
+	}
+
+	strncpy(sys_scsi_host_dir_path, sys_dev_path, PATH_MAX);
+	strncat(sys_scsi_host_dir_path, iscsi_host_dir_str,
+		PATH_MAX - iscsi_host_dir_strlen);
 
 	_good(_scandir(ctx, sys_scsi_host_dir_path, &namelist, &n), rc, out);
 
--- a/usr/idbm.c
+++ b/usr/idbm.c
@@ -27,6 +27,7 @@
 #include <errno.h>
 #include <dirent.h>
 #include <limits.h>
+#include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/file.h>
 
--- a/usr/iscsi_sysfs.c
+++ b/usr/iscsi_sysfs.c
@@ -1154,7 +1154,7 @@ int iscsi_sysfs_for_each_iface_on_host(void *data, uint32_t host_no,
 	int rc = 0, i, n;
 	struct iface_rec iface;
         char devpath[PATH_SIZE];
-        char sysfs_path[PATH_SIZE];
+        char sysfs_dev_iscsi_iface_path[PATH_SIZE];
         char id[NAME_SIZE];
 
         snprintf(id, sizeof(id), "host%u", host_no);
@@ -1164,11 +1164,11 @@ int iscsi_sysfs_for_each_iface_on_host(void *data, uint32_t host_no,
                 return ISCSI_ERR_SYSFS_LOOKUP;
         }
 
-	sprintf(sysfs_path, "/sys");
-	strlcat(sysfs_path, devpath, sizeof(sysfs_path));
-	strlcat(sysfs_path, "/iscsi_iface", sizeof(sysfs_path));
+	sprintf(sysfs_dev_iscsi_iface_path, "/sys");
+	strlcat(sysfs_dev_iscsi_iface_path, devpath, sizeof(sysfs_dev_iscsi_iface_path));
+	strlcat(sysfs_dev_iscsi_iface_path, "/iscsi_iface", sizeof(sysfs_dev_iscsi_iface_path));
 
-	n = scandir(sysfs_path, &namelist, trans_filter, alphasort);
+	n = scandir(sysfs_dev_iscsi_iface_path, &namelist, trans_filter, alphasort);
 	if (n <= 0)
 		/* older kernels or some drivers will not have ifaces */
 		return 0;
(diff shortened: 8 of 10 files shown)


cathay4t:
@gonzoleeman New PR for converting more code to library. Please review. Thanks.

libopeniscsiusr/idbm.h
```diff
@@ -0,0 +1,46 @@
...
+#include "libopeniscsiusr/libopeniscsiusr_common.h"
+
+#define ISCSI_CONFIG_ROOT	"/etc/iscsi/"
+#define IFACE_CONFIG_DIR	ISCSI_CONFIG_ROOT"ifaces"
+
+struct __DLL_LOCAL idbm;
+
+struct idbm {
+	int		refs;
+} idbm_t;
```
tasleson: It looks like this was created with the intention of doing a `typedef`, but incomplete?  You are creating an instance of `struct ibdm` here which isn't being used.

cathay4t: I forgot to remove the `idbm_t`.
Thanks.

tasleson:
If you add **any** new functions I would bump 'Y' (second digit) in version, thus 0.2.0.  Bumping 'Z' is usually reserved for fixing bugs.

tasleson:
@cathay4t Can you elaborate on the file locking?  I would like to know the high level goal here.

tasleson:
@cathay4t I'm not sure if I've reviewed any of this before, but my initial reaction is I don't like exposing the `struct iscsi_iface *iface`.  Typically C libraries leverage opaque data types to maintain compatibility when adding to structures over time.

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+	assert(ctx != NULL);
+
+	db = ctx->db;
+
+	if (db->refs > 0) {
+		db->refs++;
+		return 0;
+	}
+
+	if (access(LOCK_DIR, F_OK) != 0) {
```
tasleson: Race condition between checking the existence of creating it if it's missing.

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+#ifndef LOCK_DIR
+#define LOCK_DIR		"/var/lock/iscsi"
+#endif
+#define LOCK_FILE		LOCK_DIR"/lock"
+#define LOCK_WRITE_FILE		LOCK_DIR"/lock.write"
+
+#define _rec_str(_key, _recs, _org, _name, _show, _n, _mod) \
+do { \
+	_recs[_n].type = TYPE_STR; \
+	strncpy(_recs[_n].name, _key, NAME_MAXVAL); \
```
tasleson: If you give the entire buffer to use for `strncpy` and the source string is >= length the destination string will not be null terminated.

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+	if (strlen((char*)_org->_name)) \
+		strncpy((char*)_recs[_n].value, (char*)_org->_name, \
+			VALUE_MAXVAL); \
```
tasleson: Same as above

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+
+struct idbm_rec {
+	int			type;
+	char			name[NAME_MAXVAL];
+	char			value[VALUE_MAXVAL];
+	void			*data;
+	int			data_len;
+	int			visible;
+	/*
+	 * TODO: make it a enum that can indicate wheter it also requires
```
tasleson: wheter -> whether

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+			       "Could not lock discovery DB: %s: %d %s",
+			       LOCK_WRITE_FILE, errno,
+			       _strerror(errno, strerr_buff));
+			return LIBISCSI_ERR_IDBM;
+		} else if (i == 0)
+			_debug(ctx, "Waiting for discovery DB lock on %s",
+			       LOCK_WRITE_FILE);
+
+		usleep(10000);
+	}
```
tasleson: So as long as you don't get errno `EEXIST`, eventually you will get the lock as you'll loop 3000 times and then continue on?

tasleson:
@cathay4t Just to be perfectly clear, this library is licensed GPL, thus only GPL applications can utilize it.

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+	_recs[_n].visible = _show; \
+	_recs[_n].can_modify = _mod; \
+	_n++; \
+} while (0)
+
+#define _rec_uint32(_key, _recs, _org, _name, _show, _n, _mod) \
+do { \
+	_recs[_n].type = TYPE_UINT32; \
+	strncpy(_recs[_n].name, _key, NAME_MAXVAL); \
+	snprintf(_recs[_n].value, VALUE_MAXVAL, "%" PRIu16, _org->_name); \
```
tasleson: `PRIu32`

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+	int rc = LIBISCSI_OK;
+	int i = 0;
+	int passwd_done = 0;
+	char passwd_len[8];
+
+	assert(ctx != NULL);
+	assert(recs != NULL);
+	assert(name != NULL);
+	assert(value != NULL);
+
```
tasleson: Will `asserts` be enabled in released code?

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+		*(value+i) = 0;
+
+		rc = _idbm_rec_update_param(ctx, recs, name, value,
+					    line_number);
+		if (rc == LIBISCSI_ERR_INVAL) {
+			_error(ctx, "config file %s invalid.", conf_path);
+			goto out;
+		} else if (rc != LIBISCSI_OK)
+		      goto out;
+	} while (line);
```
tasleson: Please supply a sample config file that this is intending to parse, thanks!

libopeniscsiusr/iface.c
```diff
@@ -251,39 +194,709 @@ int _iscsi_iface_get(struct iscsi_context *ctx, uint32_t host_id, uint32_t sid,
...
+	int errno_save = 0;
+
+	cmdline[0] = "/sbin/modprobe";
+	cmdline[1] = "-qb";
+	cmdline[2] = (char *) drv_name;
+	cmdline[3] = NULL;
+
+	pid = fork();
+	if (pid == 0) {
+		if (execv("/sbin/modprobe", cmdline) < 0) {
```
tasleson: Can we use `libkmod` instead?  It's ugly to exec binaries from libraries.

libopeniscsiusr/idbm.c
```diff
@@ -0,0 +1,699 @@
...
+
+	if (iface_name == NULL)
+		goto out;
+
+	snprintf(conf_path, PATH_MAX, "%s/%s", IFACE_CONFIG_DIR, iface_name);
+
+	*iface = calloc(1, sizeof(struct iscsi_iface));
+	_alloc_null_check(ctx, *iface, rc, out);
+
+	snprintf((*iface)->name, sizeof((*iface)->name)/sizeof(char),
```
tasleson: Would it be clearer to use constant ISCSI_MAX_IFACE_LEN instead?

tasleson: All of the `snprintf` calls are not checking the return value.  I think it would be good to at least ensure they are all null terminated.

cathay4t:
@tasleson 
> I'm not sure if I've reviewed any of this before, but my initial reaction is I don't like exposing the struct iscsi_iface *iface. Typically C libraries leverage opaque data types to maintain compatibility when adding to structures over time.

Using `struct __DLL_EXPORT iscsi_iface;` instead of `typedef` does not impact the ABI on adding new
members to the struct and it also the way `libabc` suggested. Personally I don't like typedef which hide
too much information: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#typedefs

libopeniscsiusr/idbm.c
cathay4t: That's what old code doing and I guess we have to take that race.

libopeniscsiusr/idbm.c
cathay4t: Thanks for pointing that out. Fixed by a macro `_strncpy()` setting the null terminator.

libopeniscsiusr/idbm.c
cathay4t: Thanks. Fixed.

cathay4t:
@tasleson 
> Just to be perfectly clear, this library is licensed GPL, thus only GPL applications can utilize it.

Yes. A lot code are copied from old code which is GPL v2+ license, so I assume I can only license the library in GPL. Is it?

libopeniscsiusr/idbm.c
cathay4t: Using the constant might cause human error(incorrectly chose array length), runtime size calculation does not.
According to man 3 of snprintf, it will always be null terminated:
`The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.`

libopeniscsiusr/idbm.c
cathay4t: In the repo `./etc/iface.example` would do.

libopeniscsiusr/idbm.c
cathay4t: Yes. And documented as:
`If this pointer is NULL, your program will be terminated by assert.`

For internal functions, assert could help us find the error easily at early stage.

libopeniscsiusr/idbm.c
cathay4t: Fixed. Thanks.

libopeniscsiusr/idbm.c
cathay4t: Thanks. Fixed.

cathay4t:
@tasleson 
> Can you elaborate on the file locking? I would like to know the high level goal here.

1. Create `LOCK_DIR` folder and `LOCK_FILE` file if not exists.
2. If `LOCK_WRITE_FILE` exists, wait.
3. If `LOCK_WRITE_FILE` does not exist, symbol link `LOCK_FILE` to `LOCK_WRITE_FILE`.
4. On release of lock, remove `LOCK_WRITE_FILE` symbol link.

Before we finished migrating all code in iscsid/iscsiadm to use this library, we have to keep this way.

cathay4t:
V2 patch uploaded.

libopeniscsiusr/idbm.c
tasleson: Oops, you are correct, I was thinking of `strncpy`

usr/iface.c
```diff
@@ -44,6 +44,8 @@
 #include "iscsi_err.h"
 #include "iscsi_netlink.h"
 
+#define _unwrap(x) strlen(x) ? x : UNKNOWN_VALUE
+
```
tasleson: Is rust rubbing off on you? :-)

tasleson: Any possibility of x being NULL?

tasleson:
@cathay4t I've completed my review, but honestly I could spend a lot more time looking at this amount of code.  Please run coverity and see if it can catch any subtle string processing issues.

usr/iface.c
cathay4t: Yeah. That's rust.
The pointer will not be NULL, but I still change it to:
`#define _unwrap(x) (x && strlen(x)) ? x : UNKNOWN_VALUE`

libopeniscsiusr/idbm.c
gonzoleeman: @tasleson please create a separate issue on that directory check/create race? thank you

gonzoleeman:
@tasleson coverity seems to be "down" for the last month or so?

gonzoleeman:
@cathay4t what is the current state, then, of this pull request?

cathay4t:
@gonzoleeman I am trying to create a rpm with this patch for our internal coverity scan tool. Will update the result soon.

cathay4t:
@gonzoleeman I fixed all the issues found by covscan 7.3.1 in `libopeniscsiusr`. Ready for review.

gonzoleeman:
I had to manually merge libopeniscsiusr/sysfs.c. Please be sure I didn't loose any of your changes.

And thank you for your work!

libopeniscsiusr/idbm.c
tasleson: @gonzoleeman As requested: https://github.com/open-iscsi/open-iscsi/issues/95

cathay4t:
@gonzoleeman 
> I had to manually merge libopeniscsiusr/sysfs.c. Please be sure I didn't loose any of your changes.

The only changes is below which is OK for me:

```patch
--- /tmp/sysfs.c_old    2018-03-23 17:59:40.039648632 +0800
+++ libopeniscsiusr/sysfs.c     2018-03-23 17:58:12.770044835 +0800
@@ -325,6 +325,8 @@
        struct dirent **namelist = NULL;
        int n = 0;
        const char *host_id_str = NULL;
+       const char iscsi_host_dir_str[] = "/iscsi_host/";
+       const unsigned int iscsi_host_dir_strlen = strlen(iscsi_host_dir_str);
 
        assert(ctx != NULL);
        assert(sid != 0);
@@ -339,8 +341,16 @@
                                 _SYSFS_DEV_CLASS_ISCSI_SESSION, sys_dev_path),
              rc, out);
 
-       snprintf(sys_scsi_host_dir_path, PATH_MAX, "%s/iscsi_host/",
-                sys_dev_path);
+       if ((strlen(sys_dev_path) + iscsi_host_dir_strlen) >= PATH_MAX) {
+               rc = LIBISCSI_ERR_SYSFS_LOOKUP;
+               _error(ctx, "Pathname too long: %s%s",
+                      sys_dev_path, iscsi_host_dir_str);
+               goto out;
+       }
+
+       strncpy(sys_scsi_host_dir_path, sys_dev_path, PATH_MAX);
+       strncat(sys_scsi_host_dir_path, iscsi_host_dir_str,
+               PATH_MAX - iscsi_host_dir_strlen);
 
        _good(_scandir(ctx, sys_scsi_host_dir_path, &namelist, &n), rc, out);
```
Thank you.

PR #96, opened 2018-04-11
libopeniscsiusr: Full iscsi node query support.

cathay4t:
* New API functions:
        * iscsi_nodes_get()
        * iscsi_nodes_free()
        * And other node property query functions.
    
     * Replaced the node query functions used by `iscsiadm -m node` command
       which then generate identical output like old one.
    
     * Tested for back compatibility on old style node file:
            `/etc/iscsi/nodes/<target_name>/<address>,<port>`
       But only by renaming existing node file as I failed to find any
       version of iscsiadm are generating that style of node file.
    
     * API version increased to 0.2.0 due to API addition.


cathay4t:
Forgot to mention, I ran the covscan and fixed the changed codes.

gonzoleeman:
I'm sorry I haven't looked at this yet. I will look review it ASAP.

usr/iscsiadm.c
```diff
@@ -675,26 +676,122 @@ static int login_portals(struct node_rec *pattern_rec)
...
-	case 1:
-		memset(&tmp_rec, 0, sizeof(node_rec_t));
-		rc = for_each_matched_rec(rec, &tmp_rec,
-					  idbm_print_node_and_iface_tree);
-		break;
-	default:
+	if ((info_level != 0) && (info_level != -1) && (info_level != 1)) {
 		log_error("Invalid info level %d. Try 0 or 1.", info_level);
 		rc = ISCSI_ERR_INVAL;
+		goto out;
```
gonzoleeman: You are jumping to "out" here, but that frees the nodes, which we have not allocatged.

usr/iscsiadm.c
```diff
@@ -2560,18 +2657,17 @@ static int verify_node_params(struct list_head *params, struct node_rec *rec)
...
-static int exec_node_op(int op, int do_login, int do_logout,
-			int do_show, int do_rescan, int do_stats,
+static int exec_node_op(struct iscsi_context *ctx, int op, int do_login,
+			int do_logout, int do_show, int do_rescan, int do_stats,
 			int info_level, struct node_rec *rec,
 			struct list_head *params)
 {
 	int rc = 0;
 
-	if (rec)
```
gonzoleeman: The rec value can be null, so you can't remove this check.

usr/iscsiadm.c
```diff
@@ -699,21 +708,22 @@ static void print_nodes_tree(struct iscsi_node **nodes, uint32_t node_count)
...
-			printf("\tPortal: %s,%d\n",
+			printf("%s\tPortal: %s,%d\n", prefix,
 			       iscsi_node_portal_get(cur_node),
 			       iscsi_node_tpgt_get(cur_node));
 		if ((pre_node == NULL) ||
 		    (strcmp(iscsi_node_iface_name_get(cur_node),
 			    iscsi_node_iface_name_get(pre_node)) != 0))
-			printf("\t\tIface Name: %s\n",
-			       iscsi_node_iface_name_get(cur_node));
+			if (! iface_mode)
```
gonzoleeman: You generally do not put a space between the bang and the variable name.

usr/iscsiadm.c
```diff
@@ -735,7 +745,7 @@ static int print_nodes(struct iscsi_context *ctx, int info_level)
 		goto out;
 
 	if (info_level == 1)
-		print_nodes_tree(nodes, node_count);
+		print_nodes_tree(nodes, node_count, false /* not iface mode */);
```
gonzoleeman: Comments like this, i.e. saying what variable is being set, are fine for development but do not belong in production code IMHO.

usr/iscsiadm.c
cathay4t: `iscsi_nodes_free()` will do nothing against NULL pointer which is already mentioned in API document.

usr/iscsiadm.c
cathay4t: Thanks. Then I will fix the NULL deference at
```
	if ((!do_login && !do_logout && op == OP_NOOP) &&
	    (!strlen(rec->name) && !strlen(rec->conn[0].address) &&
	     !strlen(rec->iface.name))) {
```

usr/iscsiadm.c
cathay4t: Just save some time in the future when I look back on that line of code wondering what does `false` mean.
I you mind, I may change to enum which hold both meaning and value.

usr/iscsiadm.c
cathay4t: Sure. Will fix that.

cathay4t:
@gonzoleeman Have time for V2 patch set review? Thank you.

cathay4t:
@gonzoleeman Have time for V2 patch set review? Thank you.

gonzoleeman:
Reviewed and approved.

PR #100, opened 2018-04-16
Allow a host_id value of zero.

gonzoleeman:
In libopeniscsiusr, when getting iface from
sysfs, allow a host_id of zero, since the
host_id starts at zero. The fact that it
is a uint32_t is enough of a limitation.


gonzoleeman:
Any comments @cathay4t ???

PR #103, opened 2018-05-15
Reduce delays to improve iscsi boot performance

cathy-zhou:
iscsi boot performance can be improved by sleep() with finer
grained interval and exponential backoff.

Signed-off-by: Cathy Zhou <Cathy.Zhou@Oracle.COM>


gonzoleeman:
Thank you!

PR #106, opened 2018-05-30
Fix installation of libopeniscsiusr symlinks

phmccarty:
The install(1) command dereferences symlinks when copying to the target location, so use an alternate approach to create symlinks from scratch instead.


gonzoleeman:
I'm not sure I understand the issue here. I'll have to look this over more.

cleech:
Anyone want to autotool this up instead?  Or is it just me?

gonzoleeman:
On Jun 13, 2018, at 12:13 PM, Chris Leech <notifications@github.com> wrote:
> 
> Anyone want to autotool this up instead? Or is it just me?
> 
> —
> 

I have no idea what autotool is, but if it automates creating the library links I’m all for it.
— 
Lee

cleech:
> I have no idea what autotool is, but if it automates creating the library links I’m all for it.

Just the collective name for the automake / autoconf / libtool suite.  Has it's problems, but all build systems do, and I don't like trying to figure out these little details to maintain custom makefiles.

cleech:
Merging this to fix creating libopeniscsiusr.so and libopeniscsiusr.so.0 as symlinks to the fully versioned file instead of making new copies.

Thanks.

PR #107, opened 2018-06-02
Fix build with gcc8 and musl

kraj:


libopeniscsiusr/iface.c
```diff
@@ -489,7 +489,8 @@ static int _fill_hw_iface_from_sys(struct iscsi_context *ctx,
 
 	sysfs_iface_dir_path = malloc(PATH_MAX);
 	_alloc_null_check(ctx, sysfs_iface_dir_path, rc, out);
-	snprintf(sysfs_iface_dir_path, PATH_MAX, "%s/%s",
+	sysfs_iface_dir_path[PATH_MAX+strlen(_ISCSI_SYS_IFACE_DIR)] = '\0';
```
cleech: Isn't this corrupting memory right here?  sysfs_iface_dir_path is malloc(PATH_MAX), and this is zeroing a byte at PATH_MAX+strlen ...

All the PATH_MAX snprintf stuff is ugly, I think I'll look at replacing it all with asprintf if there are no objections to that.

kraj: I think you are right although, this also points the problem gcc8 has found is valid. I agree if we can reimplement it that would be cleaner. Otherwise, we can also change malloc to consider PATH_MAX+strlen(_ISCSI_SYS_IFACE_DIR)

cleech:
This has been resolved a different way with PR #110

Thank you for the suggestions, and let us know if you run into any future MUSL build issues (I have tested building the code I just merged against MUSL, but it's not something I'm going to be looking at all the time)

libopeniscsiusr/iface.c
gonzoleeman: I agree that replacing snprintf with asprintf() (almost universally) sounds excellent.

PR #110, opened 2018-06-13
Libopeniscsiusr fixes

cleech:
Fixes for iface handling regressions after merging the libopeniscsiusr code.
Found while testing under a variety of iSCSI HBAs (be2iscsi, qla4xxx, bnx2i, qedi)

### DIFF
--- a/libopeniscsiusr/iface.c
+++ b/libopeniscsiusr/iface.c
@@ -89,14 +89,14 @@ _iscsi_getter_func_gen(iscsi_iface, port_speed, const char *);
 _iscsi_getter_func_gen(iscsi_iface, name, const char *);
 
 int _iscsi_iface_get_from_sysfs(struct iscsi_context *ctx, uint32_t host_id,
-				uint32_t sid, struct iscsi_iface **iface)
+				uint32_t sid, char *iface_kern_id,
+				struct iscsi_iface **iface)
 {
 	int rc = LIBISCSI_OK;
 	char *sysfs_se_dir_path = NULL;
 	char *sysfs_sh_dir_path = NULL;
 	char *sysfs_scsi_host_dir_path = NULL;
 	char *sysfs_iface_dir_path = NULL;
-	char iface_kern_id[PATH_MAX];
 	char proc_name[ISCSI_TRANSPORT_NAME_MAXLEN];
 	struct iscsi_iface **ifaces = NULL;
 	uint32_t iface_count = 0;
@@ -195,8 +195,7 @@ int _iscsi_iface_get_from_sysfs(struct iscsi_context *ctx, uint32_t host_id,
 				(*iface)->name,
 				sizeof((*iface)->name)/sizeof(char), "");
 
-	rc = _iscsi_iface_kern_id_of_host_id(ctx, host_id, iface_kern_id);
-	if (rc == LIBISCSI_OK) {
+	if (iface_kern_id != NULL) {
 		_good(_fill_hw_iface_from_sys(ctx, *iface, iface_kern_id),
 		      rc, out);
 	} else {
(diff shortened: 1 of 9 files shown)


cleech:
@cathay4t do you want to take a look at these before I merge them?

libopeniscsiusr/sysfs.c
```diff
@@ -454,8 +455,10 @@ bool _iscsi_transport_is_loaded(const char *transport_name)
 	return false;
 }
 
-int _iscsi_iface_kern_id_of_host_id(struct iscsi_context *ctx,
-				    uint32_t host_id, char *iface_kern_id)
+int _iscsi_iface_kern_ids_of_host_id(struct iscsi_context *ctx,
+				    uint32_t host_id,
+				    char ***iface_kern_ids,
+				    uint32_t *iface_count)
```
cathay4t: @cleech Do you have any example or kernel code indicate their might be multiple ifaces binding to single scsi host?

libopeniscsiusr/sysfs.c
```diff
@@ -485,24 +488,32 @@ int _iscsi_iface_kern_id_of_host_id(struct iscsi_context *ctx,
 	_good(_scandir(ctx, sysfs_iface_path, &namelist, &n), rc, out);
 
 	if (n == 0) {
-		rc = LIBISCSI_ERR_SYSFS_LOOKUP;
+		/* this is OK, and needed for transport drivers like
```
cathay4t: There is is a whitespace in `this is OK, and needed for transport drivers like `

libopeniscsiusr/iface.c
```diff
@@ -268,6 +267,52 @@ int _iscsi_iface_get_from_sysfs(struct iscsi_context *ctx, uint32_t host_id,
...
+			_good(_iscsi_iface_get_from_sysfs(ctx, host_id, 0,
+					iface_kern_ids[i], &(*ifaces)[i]), rc, out);
+		}
+	} else {
+		/* if there's no iface exported in sysfs,
+		 * we should still be able to create one record per host */
+		*ifaces = (struct iscsi_iface **) calloc(1, sizeof(struct iscsi_iface *));
+		_alloc_null_check(ctx, *ifaces, rc, out);
+		_good(_iscsi_iface_get_from_sysfs(ctx, host_id, 0, NULL, &(*ifaces)[0]), rc, out);
+		*iface_count = 1;
```
cathay4t: Memory leak here.
If `_iscsi_iface_get_from_sysfs()` failed, `iface_count` is `0`, the memory hold by `*ifaces` will not be freed.

cathay4t:
@cleech Could you include this patch also to enable compile on GCC 8?

```patch
diff --git a/libopeniscsiusr/Makefile b/libopeniscsiusr/Makefile
index deddf3f..3c79e13 100644
--- a/libopeniscsiusr/Makefile
+++ b/libopeniscsiusr/Makefile
@@ -42,7 +42,8 @@ EXTRA_MAN_FILES = libopeniscsiusr.h.3
 OBJS = context.o misc.o session.o sysfs.o iface.o idbm.o node.o default.o
 
 CFLAGS ?= -O2 -g
-CFLAGS += -Wall -Werror -Wextra -fvisibility=hidden -fPIC
+CFLAGS += -Wall -Werror -Wextra -fvisibility=hidden -fPIC \
+         -Wno-stringop-truncation
 
 LIBADD =
 

```

libopeniscsiusr/iface.c
```diff
@@ -268,6 +267,52 @@ int _iscsi_iface_get_from_sysfs(struct iscsi_context *ctx, uint32_t host_id,
...
+	*ifaces = NULL;
+	*iface_count = 0;
+
+	_good(_iscsi_iface_kern_ids_of_host_id(ctx, host_id, &iface_kern_ids, iface_count),
+	      rc, out);
+	if (*iface_count > 0) {
+		*ifaces = (struct iscsi_iface **) calloc(*iface_count,
+							 sizeof(struct iscsi_iface *));
+		_alloc_null_check(ctx, *ifaces, rc, out);
+		for (uint32_t i = 0; i < *iface_count; i++) {
```
cathay4t: Got compile failure on RHEL 7.5:
```
sysfs.c: In function ‘_iscsi_iface_kern_ids_of_host_id’:
sysfs.c:502:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for (uint32_t i = 0; i < *iface_count; i++) {
  ^
sysfs.c:502:2: note: use option -std=c99 or -std=gnu99 to compile your code
sysfs.c:510:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (uint32_t i = 0; i < *iface_count; i++ ) {
   ^
cc1: all warnings being treated as errors
make: *** [sysfs.o] Error 1
```

cathay4t:
I will create new PR for gcc8 compiling issue.

cathay4t:
GCC 8 compile issue fixed in #112.

libopeniscsiusr/iface.c
cleech: good find, I'll move setting *iface_count = 1 up a line, so it's between the _alloc_null_check and the call to _iscsi_iface_get_from_sysfs

libopeniscsiusr/iface.c
cleech: This brings up a larger conversation about language standards, compiler extensions, and coding style.  We should probably make some choices about what to allow, and then get the CFLAGS set to match.  I'm not too hung up on loop initial declarations, but in general I think allowing at least C99 can result in much cleaner code.

Related but separate, I was working up a patch to use asprintf instead of snprintf to the stack and removing all PATH_MAX buffers.  asprintf is non-standard, but we only care about Linux and it's supported in glibc and musl.

In the process, I also started using the gcc __attribute__((__cleanup__(f))) to make it obvious which allocations are local to the function and get the compiler to emit code to always free them as they leave scope.  I may have gotten use to the idea when working on systemd code, and think it makes C much nicer to deal with (with some helper cleanup functions and macros to make the attribute syntax easier to look at).

libopeniscsiusr/sysfs.c
cleech: Both be2iscsi and qla4xxx (so, the only two users of the iscsi_create_iface API) will create both IPv4 and IPv6 ifaces depending on the firmware setting on the card.  I think you do need to enable IPv6 in the HBA firmware, because I know my qla4xxx test box was creating 3-per-host at one time and right now it's only doing 1.

cleech:
updated PR after fixing for review comments

added asprintf conversion from snprintf to deal with truncation issues

I've included just the strncpy => memset portion of #112 here as I prefer the asprintf solution over increasing stack buffer for the other part.  I'll close #112

PR #112, opened 2018-06-14
libopeniscsiusr: Fix compile error on GCC 8.

cathay4t:
Fix two compile warnings(will be treated as error):
 * `strncpy` for stringop-truncation warning:
  Use memcpy instead and manually add trailing '\0'.
 * `snprintf` for format-truncation warning:
  Double the buffer size.

### DIFF
--- a/libopeniscsiusr/iface.c
+++ b/libopeniscsiusr/iface.c
@@ -487,9 +487,9 @@ static int _fill_hw_iface_from_sys(struct iscsi_context *ctx,
 	assert(iface != NULL);
 	assert(iface_kern_id != NULL);
 
-	sysfs_iface_dir_path = malloc(PATH_MAX);
+	sysfs_iface_dir_path = malloc(PATH_MAX * 2);
 	_alloc_null_check(ctx, sysfs_iface_dir_path, rc, out);
-	snprintf(sysfs_iface_dir_path, PATH_MAX, "%s/%s",
+	snprintf(sysfs_iface_dir_path, PATH_MAX * 2, "%s/%s",
 		 _ISCSI_SYS_IFACE_DIR, iface_kern_id);
 
 	_good(_sysfs_prop_get_str(ctx, sysfs_iface_dir_path,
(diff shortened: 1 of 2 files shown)


cleech:
I really don't like making these huge stack buffers to protect against truncation warnings.  PATH_MAX isn't real (filesystems don't do that, they may have filename limits) but it provides a somewhat sane upper limit for allocations.  But then we just double it for concatenation, now that's an 8k stack buffer.

I'd rather switch to heap allocations and get these off the stack entirely.  I've got a patch ready to switch everything over to asprintf.

The basic pattern looks like this:
```
-    char path[PATH_MAX];
+    char *path = NULL;

-    snprintf(path, PATH_MAX, "%s/%s", PATH_DIR, dynamic_part);
+    _good(_asprintf(&path, "%s/%s", PATH_DIR, dynamic_part), rc, out);
 out:
+    free(path);
```
In some cases the code is already allocating with malloc
```
     char *path = NULL;

-    path = malloc(PATH_MAX);
-    _alloc_null_check(ctx, path, rc, out);
-    snprintf(path, PATH_MAX, "%s/%s", PATH_DIR, dynamic_part);
+    _good(_asprintf(&path, "%s/%s", PATH_DIR, dynamic_part), rc, out);
 out:
     free(path)
```
where ```_asprintf``` is a return code wrapper to make it work with ```_good```
```
#define _asprintf(...) (asprintf(__VA_ARGS__) == -1 ? LIBISCSI_ERR_NOMEM : LIBISCSI_OK)
```
So ```_good(_asprintf))``` covers the allocation check of ```_alloc_null_check``` and is safe against truncation, which we aren't checking for with ```snprintf```

If we want to use ```__attribute__((cleanup)) ``` to take care of these sort of allocation, then the ```free``` calls get hidden as cleanup calls inserted by the compiler.
```
- char *path = NULL;
+ _cleanup_free_ char *path = NULL;

-    path = malloc(PATH_MAX);
-    _alloc_null_check(ctx, path, rc, out);
-    snprintf(path, PATH_MAX, "%s/%s", PATH_DIR, dynamic_part);
+    _good(_asprintf(&path, "%s/%s", PATH_DIR, dynamic_part), rc, out);
 out:
-    free(path)
```
and we have some utility inlines and macros to make using the attribute saner
```
#define _cleanup_(f) __attribute__((cleanup(f)))

static inline void _freep(void *p) {
    free(*(void **)p);
}
#define _cleanup_free_ _cleanup_(_freep)
```

cleech:
I've put the strncpy => memcpy portion of this change in with an updated #110 
For the PATH_MAX stack buffers, I prefer to switch to asprintf as detailed above. 
A change for that is also in an updated #110

Thanks

PR #114, opened 2018-06-20
iscsiuio: Add inter-host mutex while doing xmit

njavali:
This avoids the netlink buffer corruption when more than one host
try to xmit packet at the same time.

Signed-off-by: Manish Rangankar <manish.rangankar@cavium.com>

### DIFF
--- a/iscsiuio/src/unix/libs/qedi.c
+++ b/iscsiuio/src/unix/libs/qedi.c
@@ -75,6 +75,8 @@
 
 extern int nl_sock;
 
+static pthread_mutex_t host_mutex = PTHREAD_MUTEX_INITIALIZER;
+
 /*  Foward struct declarations */
 struct nic_ops qedi_op;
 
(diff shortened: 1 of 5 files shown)


cleech:
That's a lot of debug logging.  Which is OK, but out of the 52 lines added only 4 look to be the actual change described.

iscsiuio/src/unix/libs/qedi.c
```diff
@@ -58,6 +58,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <sys/sysmacros.h>
+#include <semaphore.h>
```
cleech: I don't think this is needed, it's for IPC semaphores while the added code here uses a pthread mutex

gonzoleeman:
I agree with @cleech . I can live with the slightly-excessive log messages. Please remove the include of "semaphore.h" if it's unneeded, and after that I'm good with this.

njavali:
I have removed unused "semaphore.h" and updated the patch. Please consider the pull request.

Thanks,
Nilesh

PR #115, opened 2018-06-22
enable MaxOutstandingR2T negotiation

cleech:
The iscsi_tcp kernel side R2T handling is in libiscsi_tcp and is shared
with at least cxgbi3, and the drivers claim CAP_MULTI_R2T. But the
iscsid login parameter negotiation code never enabled MaxOutstandingR2T.

I've tested this against an LIO target, negotiating to 2 and 4 R2Ts.
The network traces looked good, with the initiator keeping track of
multiple outstanding R2Ts and using the correct next identifier.
Throughput of bulk write data was improved, even over low latency virtio
links.

### DIFF
--- a/usr/initiator.h
+++ b/usr/initiator.h
@@ -212,6 +212,7 @@ typedef struct iscsi_session {
 	int erl;
 	uint32_t imm_data_en;
 	uint32_t initial_r2t_en;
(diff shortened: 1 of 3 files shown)


gonzoleeman:
I'm good with this @cleech   -- please feel free to integrate.

Issue #118, opened 2018-06-29
Memory leak in usr/discovery.c: discovery_sendtargets()


zhangzujian:
In file usr/login.c lin 328:

```c
session->target_alias = malloc(size + 1);
```

*target_alias* is not freed in *iscsi_destroy_session()* nor in *iscsi_free_session()*.

gonzoleeman:
I think this will do it:
```
diff --git a/usr/discovery.c b/usr/discovery.c
index 8001af6fe452..2147deb2eda1 100644
--- a/usr/discovery.c
+++ b/usr/discovery.c
@@ -1055,6 +1055,11 @@ static void iscsi_destroy_session(struct iscsi_session *session)
                log_error("Could not safely destroy session %d (err %d)",
                          session->id, rc);
 done:
+       if (session->target_alias) {
+           free(session->target_alias);
+           session->target_alias = NULL;
+       }
+
        if (conn->socket_fd >= 0) {
                ipc->ctldev_close();
                conn->socket_fd = -1;

```

gonzoleeman:
I created pull request #123 for this.

PR #123, opened 2018-08-30
Plugging a memory leak from discovery.

gonzoleeman:
During discovery the function get_op_params_text_keys()
can allocate a target_alias, but this is never freed.

Discovered-by: zhangzujian.7@gmail.com


gonzoleeman:
See issue #118

PR #129, opened 2018-09-20
Cleanup nic nl open usage

### LINKED ISSUE #125: at the moment nic_nl_open retval is neither correct nor handled
Hi,
we were just trying to debug something and came by this code and where wondering.
The RC of nic_nl_open is defined as ["0 on success, negative on error"](https://github.com/open-iscsi/open-iscsi/blob/master/iscsiuio/src/unix/nic_nl.c#L533)

But while the code sets rc to some values and might break to the error: label, it then does
[return 0;](https://github.com/open-iscsi/open-iscsi/blob/master/iscsiuio/src/unix/nic_nl.c#L679)

I wonder if it should actually initialize `rc = 0` at the beginning and `return rc;` eventually.

Furthermore on the [call to the function](https://github.com/open-iscsi/open-iscsi/blob/master/iscsiuio/src/unix/main.c) it totally ignored the retval which should maybe lead to a bad exit of the service.

These were just the thoughts when looking at the code, thanks to [Smoser](https://github.com/smoser) for making me aware of all of this. I just wanted to file it to get your opinion on it.

So do you agree to the review?
Would you want a PR to fix it or are we missing part of the concept how this is supposed to work?

cpaelzer:
As discussed in issue #125 there are a few non critical cleanups that could be done.
Due to the way nic_nl_open is currently called from main these changes should not change execution but will help to avoid breaking things later by accident - for example by inserting code in main after the nic_nl_open call.

Fixes #125

### DIFF
--- a/iscsiuio/src/unix/main.c
+++ b/iscsiuio/src/unix/main.c
@@ -393,7 +393,8 @@ int main(int argc, char *argv[])
 		goto error;
 
 	/*  NetLink connection to listen to NETLINK_ISCSI private messages */
-	nic_nl_open();
+	if (nic_nl_open() != 0)
+		goto error;
 
 error:
 	cleanup();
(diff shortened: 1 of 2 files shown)


gonzoleeman:
Thank you!

Issue #126, opened 2018-09-19
RFC: stale session information causes errors when iscsid restarts: How to Fix?


gonzoleeman:
I mentioned this issue on the mailing list, but evidently open-iscsi lives in a split-brain universe these days, so I wanted to post it here for those that don't read the mailing list regularly.

I have an issue when iscsid restarts when there is stale session information. To reproduce:

1- create an active session between an initiator and target
2- stop the initiator daemon (iscsid)
3- disable the target. I did this with LIO by removing the portal, so that there is no way to get to the target
4- start iscsid again

when iscsid restarts, it see the stale session information. As a matter of fact, as soon as it sees this information, it tries to log back in to the target, to re-establish the connection and it's attributes. This re-logging in works just fine if the target is present, but in this case, since the target is not present, it starts retrying the login, and continues to retry every two seconds, forever.

First of all, this "retry re-logging in forever" behavior seems broken. There should be a limit. There's actually a "relogin_max" session attribute, but it appears to only be used for discovery. So that's bug number 1 in my opinion.

If we assume that bug#1 above gets fixed, then there is still going to be a window of time (minutes?) where the session information is stale, that is until it gives up retrying finally times out. The question then becomes what should session information look like during this time.

One option is to have iscsiadm return an error if called to display session information during this time. Something like "Target not connected -- trying to recover". In this case the caller would simply retry later. The second option I can see here is for iscsiadm to have a "force logout" option for session mode, which will tell iscsiadm/iscsid to end the session, no matter what, since the target is never coming back. This would allow a "fail fast" code path, in effect.

I tend to like the first option, i.e. the session ends when the retry-maximum count is reached, and any attempt to see session information before then returns an appropriate error, I am working on such a solution now, but it's slow going since I don't know the state/connection/login code very well (yet). So any suggestions are more than welcome, as are comments and kibitzing. Anyone?

gonzoleeman:
I believe my pull request fixes this issue, at least for now.

gonzoleeman:
Closing since request #131 was merged.

PR #131, opened 2018-09-26
Limit iscsid infinite session reconnect attempts at startup using a new config value

gonzoleeman:
This set of changes includes some bug fixes for existing problems, but they were problems related to the main issue, so they are fixed as well.

The main issue is that iscsid tries to reconnect "stale" sessions it finds when it starts up, and it does this forever. So we limit that, and we fix up access to the session during that time, as well.


gonzoleeman:
See issue #126

gonzoleeman:
@cleech can you comment on this? @mikechristie ?

gonzoleeman:
Having not yet heard a reply I am going to merge this in.

PR #132, opened 2018-09-27
Small cleanups

gonzoleeman:
I was tired of seeing 3 non-fatal warnings when compiling so this fixes them. Gcc8 is now much happier.


gonzoleeman:
Merging these 3 small changes in.

Issue #135, opened 2018-10-04
Race condition on creating multiple sessions to a target


gonzoleeman:
The iscsid configuration variable node.session.nr_sessions defaults to "1", meaning that iscsid should create only 1 session to a particular target. (see iscsid.conf).

So if you try to create a second session after creating a first session, you will be told:

```
iscsiadm: default: 1 session requested, but 1 already present.
iscsiadm: Could not log into all portals
```

But if you try to log into a target more than once in parallel, it's likely to succeed, e.g.:

`bash# (iscsiadm -m node -l &; iscsiadm -m node -l &)`

If you try this with your initiator and target on the same host, it will generally fail. But if you try it from a remote node, there is enough delay so that it succeeds!

This occurs because of the asynchronous nature of a login request. The sequence is:

* if current number of sessions to that target is >= nr_sessions, exit with error
* send "nr_sessions" login requests to the target

The "current number of sessions" is check in sysfs. So, if you have requested a login to a target but it has not yet responded, or if it has responded but the attributes haven't shown up yet in sysfs, then iscsid will think there is no current session and allow a second session.

In order for this to work, there is some code in iscsid:

```
         /*
         * Ensure the record's 'multiple' flag is set so __iscsi_login_portal
         * will allow multiple logins.
         */
        rec->session.multiple = 1;
```

I believe the fix for this issue is to change the code to only set this multiple flag if nr_sessions is greater than 1, and testing has shown that fixes the issue.

I will submit a pull request.

gonzoleeman:
Closing as resolved.

PR #136, opened 2018-10-04
Do not allow multiple sessions when nr_sessions=1

gonzoleeman:
If a request is made to login to a target, creating
a session, then another request is submitted so quickly
that the first one has not completed, both  requests
can succeed, despite having nr_sessions=1 configured.

Only allow multiple login requests if nr_sessions is
greater than 1.


gonzoleeman:
See Issue #135

gonzoleeman:
Have received no comments, I merged this in.

Issue #140, opened 2018-10-10
iscsid should use sd_notify() to work with systemd


gonzoleeman:
It seems like we can't get away from using systemd these days. So we might as well update iscsid to notify system when it is up and running, since, it it's current form, systemd think iscsid is ready immediately upon startup to handle iscsi requests.

gonzoleeman:
done

PR #143, opened 2018-10-16
Use sd notify for systemd

gonzoleeman:
NOTE: This REQUIRES an update to the sytemd unit service file iscsid.service. See commit comments.


gonzoleeman:
See Issue #140

PR #148, opened 2018-11-20
iscsiuio bug fixes

njavali:
Please incorporate the below iscsiuio bug fixes.

Thanks,
Nilesh

### DIFF
--- a/iscsiuio/README
+++ b/iscsiuio/README
@@ -1,6 +1,6 @@
 Iscsiuio Userspace Tool
-Version 0.7.8.4
-Feb 22, 2018
+Version 0.7.8.5
+Nov 20, 2018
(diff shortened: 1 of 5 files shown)


gonzoleeman:
On your second patch, it says it has a dependency on an upstream patch. What happens if we run the updated iscsiuio with an unpatched kernel?

njavali:
> On your second patch, it says it has a dependency on an upstream patch. What happens if we run the updated iscsiuio with an unpatched kernel?

The combination of qedi and iscsiuio fixes the overall iscsiuio packet processing. Hence in case only iscsiuio is updated with an unpatched kernel then the user may experience iSCSI initiator IP ping drop issue.

PR #154, opened 2019-02-02
Make iscsid.service a requirement.

igo95862:
This will ensure daemon is loaded before iscsi.service starts. Also in case daemon fails to load iscsi.service will not be started.


gonzoleeman:
Thank you.

PR #157, opened 2019-02-13
[WIP] Make systemd optional but enabled by default

maxice8:
This is a WIP work i cooked up in a time to make systemd optional:

- Distros that do not use systemd (like Void Linux and Alpine Linux) can now pass NO_SYSTEMD and the systemd-specific bits will be replaced by no-ops, the systemd/sd-daemon.h include will be removed.
- Distros that use systemd (RHEL/Fedora/OpenSUSE/Arch/Debian/etcetcetc) just do nothing and it will keep going as usual.

### DIFF
--- a/usr/Makefile
+++ b/usr/Makefile
@@ -35,13 +35,18 @@ endif
 PKG_CONFIG = /usr/bin/pkg-config
 
 CFLAGS ?= -O2 -g
+ifdef NO_SYSTEMD
+CFLAGS += -DNO_SYSTEMD
+endif
 WARNFLAGS ?= -Wall -Wstrict-prototypes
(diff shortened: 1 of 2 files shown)


gonzoleeman:
You may have to also play with the ISCIADM_NAMESPACE stuff .. I'm not sure. It was designed to work whether or not systemd socket activation was being used, so it may "just work" (tm)

Which brings me to my last question: have you tested this? Is it ready for me to test it?

gonzoleeman:
I'm sorry, what a space head. I forgot you had these pull requests pending, so I ended up doing almost exactly the same thing in pull request #164 

Is it okay with you if we use that one. the only difference is the format used in the Makefile.

gonzoleeman:
Closing this, as I merged #164 -- sorry it took so long to get to this!

PR #159, opened 2019-03-12
Fix node print return value when no nodes.

gonzoleeman:
When we switched to using libopeniscsiusr, the behavior of iscsiadm
when running "iscsiadm -m node" diverged with that seen when
running "iscsiadm -m node --op show", whereas before that time
(version 2.0.873) was the same for these two cases. Note that
iscsiadm return value used to be the same in both cases, i.e.
returning 21, meaning "no targets found", but now returns
21 for the "--op show" case and zero for the other.

The fix is simple: rip out the code in iscsiadm that treats
"iscsiadm -m node" as a special case, allowing the code to fall
through to its previous behavior.

### DIFF
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -681,14 +681,6 @@ static int login_portals(struct node_rec *pattern_rec)
 	return rc;
 }
 
-static void print_node_flat(struct iscsi_node *node)
-{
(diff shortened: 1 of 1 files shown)


gonzoleeman:
This fixes a regression caused by our switch to libopeniscsiusr. I will merge if no objection.

PR #160, opened 2019-03-12
Add Restart=on-failure option to iscsid.service

fredvx:
This change adds Restart=on-failure systemd service option to iscsid.service config file so that iscsid daemon is restarted on failure.  This option is particularly useful when using iscsi boot device and iscsid daemon crashes or is inadvertently killed.


gonzoleeman:
Thank you.

PR #161, opened 2019-03-26
Fix printing of node database again.

gonzoleeman:
This fixes commit ec2afb8a1c997bac706de4f70a3a59fdd82b5741,
which aimed to make the return value from "iscsiadm -m node -P[01]"
return the correct value of "21", for no targets found. But
instead it make the output of all "iscsiadm -m node -P*" act like
the output of "iscsiadm -m node --op show", which was just wrong.

Instead, revert the above-mentioned commit, then add a simple
check for "no nodes" in print_nodes().

### DIFF
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -681,6 +681,14 @@ static int login_portals(struct node_rec *pattern_rec)
 	return rc;
 }
 
+static void print_node_flat(struct iscsi_node *node)
+{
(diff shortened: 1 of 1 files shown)


gonzoleeman:
Note: the output of "iscsiadm -m node -P*" has been broken since I committed ec2afb8. This fixes it back to what it should be, plus fixes the problem that commit was designed to fix, which is return 21 when there are no Nodes in the database.

PR #162, opened 2019-03-26
Fix output of node printing for multiple paths.

gonzoleeman:
The output of "iscsiadmm -m node -P*" changed when
libopeniscsiusr was merged, and an earlier attempt was
made to clean it up (see commit 59806c99fb3f). This
commit fixes output so that. Before this fix, if one
had multiple paths to the same target, the output
of "iscsiadm -m node -P1" would be:

> Target: iqn.2008-04.net.gonzoleeman:01:disk1
>         Portal: 10.0.0.1:3260,1
>                 Iface Name: default
> Target: iqn.2008-04.net.gonzoleeman:01:disk1
>         Portal: 10.0.1.1:3260,2
>                 Iface Name: default

But this commit fixes the output to be that of iscsiadm
before the libopeniscsiusr merge, i.e.:

> Target: iqn.2008-04.net.gonzoleeman:01:disk1
>         Portal: 10.0.0.1:3260,1
>                 Iface Name: default
>         Portal: 10.0.1.1:3260,2
>                 Iface Name: default

### DIFF
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -689,13 +689,12 @@ static void print_node_flat(struct iscsi_node *node)
 		       iscsi_node_target_name_get(node));
 }
 
-// The 'iface_mode' argument is only used for command
-//	`iscsiadm -m iface -P 1`
 static void print_nodes_tree(struct iscsi_node **nodes, uint32_t node_count,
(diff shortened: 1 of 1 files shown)


gonzoleeman:
I _really_ need to get the test suite working for open-iscsi so little stuff like this doesn't happen.

PR #163, opened 2019-05-14
Fix systemd startup etc

gonzoleeman:
This group of commits brings iscsiuio up to date with iscsid, in that it adds support for systemd at startup time, but in this case it can be ifdefed out, for those cases when systemd is not available.

It also adds example systemd service and socket files for iscsiuio, and it cleans up the error messages in iscsiuio, which didn't all have newlines, and were being dumped to stdout.

Lastly, for those still wishing to use daemon (non-foreground) mode for iscsiuio, I add a handshake between parent and child when becoming a daemon, so that the parent does not exit until the child is actually ready to do business.

### DIFF
--- a/doc/iscsid.8
+++ b/doc/iscsid.8
@@ -40,7 +40,7 @@ do not write a process ID file.
 .TP
 .BI [-p|--pid=]\fIpid\-file\fP
 write process ID to \fIpid\-file\fR rather than the default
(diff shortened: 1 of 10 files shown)


gonzoleeman:
Note: this issue was originally found by a co-worker, but I won't post his name (thanks GDPR)

Issue #63, opened 2017-08-24
iscsiadm sometimes does not show attached device (iscsiadm -m session) but device is really exist and attached


morden2k:
Hi colleagues. 

I have some troubles with `iscsiadm`. `iscsiadm` sometimes creates LUN but does not show attached disk. My device is really exist.

Login:

    >sudo -s iscsiadm --mode node --portal 192.168.1.86 --targetname iqn.2010-06.com.nutanix:vg-restore-99079f45-f62d-4df5-8955-d3f6ffa27879 --login 2>&1

    Logging in to [iface: default, target: iqn.2010-06.com.nutanix:vg-restore-99079f45-f62d-4df5-8955-d3f6ffa27879, portal: 192.168.1.86,3260] (multiple)
    Login to [iface: default, target: iqn.2010-06.com.nutanix:vg-restore-99079f45-f62d-4df5-8955-d3f6ffa27879, portal: 192.168.1.86,3260] successful.

Check result:

    >sudo -s iscsiadm -m session -P 3 2>&1

    Target: iqn.2010-06.com.nutanix:vg-backup-3943a3b9-3745-4dd7-ba0e-72f25b4a417b
        Current Portal: 192.168.1.86:3260,1
        Persistent Portal: 192.168.1.86:3260,1
            **********
            Interface:
            **********
            Iface Name: default
            Iface Transport: tcp
            Iface Initiatorname: iqn.1993-08.org.debian:01:5bb3c3d5f39
            Iface IPaddress: 192.168.0.43
            Iface HWaddress: <empty>
            Iface Netdev: <empty>
            SID: 10
            iSCSI Connection State: LOGGED IN
            iSCSI Session State: LOGGED_IN
            Internal iscsid Session State: NO CHANGE
            *********
            Timeouts:
            *********
            Recovery Timeout: 120
            Target Reset Timeout: 30
            LUN Reset Timeout: 30
            Abort Timeout: 15
            *****
            CHAP:
            *****
            username: <empty>
            password: ********
            username_in: <empty>
            password_in: ********
            ************************
            Negotiated iSCSI params:
            ************************
            HeaderDigest: None
            DataDigest: None
            MaxRecvDataSegmentLength: 262144
            MaxXmitDataSegmentLength: 1048576
            FirstBurstLength: 262144
            MaxBurstLength: 16776192
            ImmediateData: Yes
            InitialR2T: Yes
            MaxOutstandingR2T: 1
            ************************
[...]

PR #164, opened 2019-05-14
Make iscsid systemd usage optional

gonzoleeman:
You can compile without system now by using something
like:

	make OPTFLAGS="-DNO_SYSTEMD ..." NO_SYSTEMD=1

This will skip systemd code for iscsid and iscsiuio.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,11 @@ ifneq (,$(CFLAGS))
 export CFLAGS
 endif
 
+# export systemd disablement if set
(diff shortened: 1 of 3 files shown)


gonzoleeman:
Note: this commit somewhat depends on pull request #63 in that iscsiuio does not care about systemd unless that pull request is merged.

gonzoleeman:
Note: this a duplicate of #157 but I like the Makefile changes in this one a bit more. Sorry for the duplication of effort.

Issue #155, opened 2019-02-11
iscsiadm discover command pegging CPU at 100%


vishnuitta:
`iscsiadm -m discoverydb -t st -p 127.0.0.1:3260 --discover` command gets into infinite loop pegging iscsiadm CPU usage at 100%.
This can happen when response is NOT received in less than configured active timeout value for 'SendTargets=All' request for all its retries.

Below logs are taken with 'active timeout' setting at 20 seconds and modified iSCSI target which delays in responding for 'SendTargets=All' request.
In every retry, 'Discovery' request is served, and, 'SendTargets' is timedout.

```
root@localhost:# iscsiadm -d 10 -m discoverydb -t st -p 127.0.0.1:3260 --discover
iscsiadm: ip 127.0.0.1, port 3260, tgpt -1
iscsiadm: Max file limits 1024 1048576

iscsiadm: Looking for config file /etc/iscsi/send_targets/127.0.0.1,3260

iscsiadm: Looking for config file /etc/iscsi/send_targets/127.0.0.1,3260 config st_config.
iscsiadm: Could not stat /etc/iscsi/send_targets/127.0.0.1,3260 err 2.
iscsiadm: Could not open /etc/iscsi/send_targets/127.0.0.1,3260: No such file or directory

iscsiadm: Discovery record [127.0.0.1,3260] not found!
iscsiadm: updating defaults from '/etc/iscsi/iscsid.conf'
iscsiadm: updated 'discovery.sendtargets.iscsi.MaxRecvDataSegmentLength', '32768' => '32768'
iscsiadm: updated 'discovery.sendtargets.timeo.active_timeout', '30' => '20'
iscsiadm: updated 'node.startup', 'manual' => 'manual'
iscsiadm: updated 'node.leading_login', 'No' => 'No'
iscsiadm: updated 'node.session.timeo.replacement_timeout', '120' => '120'
iscsiadm: updated 'node.conn[0].timeo.login_timeout', '30' => '15'
iscsiadm: updated 'node.conn[0].timeo.logout_timeout', '15' => '15'
iscsiadm: updated 'node.conn[0].timeo.noop_out_interval', '5' => '5'
iscsiadm: updated 'node.conn[0].timeo.noop_out_timeout', '5' => '5'
iscsiadm: updated 'node.session.err_timeo.abort_timeout', '15' => '15'
iscsiadm: updated 'node.session.err_timeo.lu_reset_timeout', '30' => '30'
[...]

PR #165, opened 2019-05-14
Fix possible discovery hang when timing out

gonzoleeman:
If session reopening failed during discovery, the CPU
could peg at 100% because the code that gives up when
tretires are exhausted was not setting an error
return value.


gonzoleeman:
See issue #155

Issue #24, opened 2016-08-11
RFE: iscsi CHAP mode is not FIPS-compliant


gonzoleeman:
I'd like a note added in the README about the fact that the iSCSI CHAP mode is not fips-compliant.

This is not an open-iscsi issue but an iSCSI Standard issue. CHAP is just not secure enough for fips. And the iSCSI standard is not likely to change anytime soon.

Reference to the fips standard: http://csrc.nist.gov/publications/fips/fips140-2/fips1402.pdf

gonzoleeman:
Marked as an enhancement request.

gonzoleeman:
See Pull Request #170

gonzoleeman:
This may be fixed now, but I will wait until I can verify that before closing.

gonzoleeman:
I believe this is fixed now, with new encryption algorithm support, but I've yet to test it.

gonzoleeman:
I believe this is done, now, with recent enhancements using openssl.

PR #170, opened 2019-07-24
RFC: CHAP SHA1, SHA3-256 via OpenSSL's libcrypto

cleech:
Extended CHAP authentication modes using newer hash digest functions are going to be proposed to the linux-scsi development list for the kernel target.  I've prepared this initiator side implementation to assist in the review and testing of that proposal.

I expect most of the discussion to take place on linux-scsi.  Once there is a target that adopts an authentication mode beyond CHAP-MD5, that would be the time to merge initiator support in Open-iSCSI.


gonzoleeman:
Wonderful! I've had something like this on my todo list for a long time, but of course it has to be coordinated with some target. Nice! I can't wait to try it.

maurizio-lombardi:
Here are the target patches I am preparing, they introduce support for SHA1 and SHA3-256

https://github.com/maurizio-lombardi/linux/commits/upstream_chapfips

ddiss:
nice to see iSCSI auth making it's way into the 21st century :)
Are you aware of any other target or initiator implementations which support these digests?

cleech:
I'm not aware of any iSCSI implementation that supports any authentication other than CHAP/MD5.  We're proposing this as a simple step forward that doesn't address other potential issues with CHAP, as moving to an entirely new authentication protocol would obviously be a bigger step (and should probably be driven by someone with a more direct interest in the iSCSI storage target space).

gonzoleeman:
Chris, if it's okay with you, I believe we should merge these changes in.

David, I have seen a set of patches submitted to linux-scsi and target-devel by Maurizio that would add this support to LIO.

gonzoleeman:
BTW, I've been asked to tag a new version of open-iscsi, so it'd be great to get this in first IMHO.

cleech:
> Chris, if it's okay with you, I believe we should merge these changes in.
> 
> David, I have seen a set of patches submitted to linux-scsi and target-devel by Maurizio that would add this support to LIO.

I'd like to see what the reception to the target patch set is, if there's objections to extending CHAP there it'd be pointless to merge it here (although we could always just remove it).

gonzoleeman:
Chris, I understand your view. But unless we merge this, there will be nothing to test the target code out with.

And I suspect the upstream response will be similar to the response here, i.e. crickets. But I'm in no hurry.

gonzoleeman:
So I am you suggested waiting, Chris, as the discussion on target-devel seems to imply there may be more to it than adding SHA3-256.

gonzoleeman:
See Issue #24

gonzoleeman:
Note: the other end of this discussion, i.e. the target side, has seen some lively discussion, particularly with respect to which algorithms to support, and ensuring the result is registered as a legal part of the protocol.

Because of that, it may take a while, so I will not be waiting for this stuff to tag a new version of open-iscsi.

But I'm open to adding it as _experimental_, if there was some way to do that. Perhaps a new (temporary?) branch, or a build option in the sources? At the very least, I suspect there will be those that want to to have something with which to talk to test targets implementing new encryption algorithms. Any suggestions/ideas?

maurizio-lombardi:
An experimental branch looks like a good compromise IMO.
About the target side: I am soon going to contact IETF to see how we can proceed to have this approved.

cleech:
I've updated the PR.  IANA assignments for this are done (https://www.iana.org/assignments/ppp-numbers/ppp-numbers.xhtml#ppp-numbers-9) so this is now a standards based enhancement and not an experimental extension ;-)

I think this is ready to merge now, matching target side kernel code should catch up soon.

gonzoleeman:
I think that's great! Would you like to merge it, or would you prefer I do it?

Also, I'm thinking we should tag a new version, even though I just tagged 2.0.878. Since this is somewhat of a major change, perhaps bumping to 2.1 is in order?

PR #176, opened 2019-10-23
setup Travis-CI builds and Coverity scans

cleech:
Configure Travis-CI for automated build tests.
Hopefully we can extend this to basic functional as well.

Also sets up Coverity scanning from the Travis build.
Synopsys gives Open Source projects a limited number of free Coverity
scans each day, so this is currently configured to only scan when code
is pushed to a "coverity_scan" branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: c
+compiler: gcc
+os: linux
+dist: bionic
+
+env:
+  global:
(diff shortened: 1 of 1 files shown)


gonzoleeman:
Perhaps we should automate pushing to this new branch for each commit, since we don't have that many?

cleech:
We can change it to go ahead and scan on master if wanted.
I'd like to be able to scan PRs before we merge them.

gonzoleeman:
Perhaps we need a "for-next" branch? Perhaps a manual (for now) merge from there to master at appropriate times?

There are a set of open-iscsi tests that wouldn't take too much work to get going. That has been on my todo list for a couple of years. :(

PR #178, opened 2019-10-31
first batch of changes to address Coverity scan issues

cleech:
This closes out 31 reported "high impact" issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ dist: bionic
 
 env:
   global:
+  # This is an excrypted setting of COVERITY_SCAN_TOKEN=<token>
+  # Travis-CI has the private key to decrypt this and set the environment
+  # variable with the token needed for Coverity API access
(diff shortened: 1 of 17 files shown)


gonzoleeman:
I am testing this, and it looks good so far. It seems to fix an issue -- one that I spent the better part of a day fixing -- where normal CHAP wasn't working correctly when compiled with gcc's "-fstack-protector". :O

gonzoleeman:
Note: testing with regular (non-authenticated) connections works, and with normal CHAP. I have yet to test (before this branch or on this branch) the new authentication options.

cleech:
> I am testing this, and it looks good so far. It seems to fix an issue -- one that I spent the better part of a day fixing -- where normal CHAP wasn't working correctly when compiled with gcc's "-fstack-protector". :O

I was pretty embarrassed when I saw the scan issue with the newly modified auth code, missed making a stack array bigger when we switched from adding 1 new digest to 2.

Thanks for taking a look at all this, I'll get some additional tests run myself.

gonzoleeman:
LOL. No need to be embarrassed! That was a bunch of changes.

Also, I'd like to test the newer authentication algorithms, but I don't know how! I thought LIO/targetcli-fb were already updated to support the newer algorithms, but I don't see the code! Plus I'm not sure how to change my algorithm preference it in the initiator. Any hints?

cleech:
> Also, I'd like to test the newer authentication algorithms, but I don't know how! I thought LIO/targetcli-fb were already updated to support the newer algorithms, but I don't see the code! Plus I'm not sure how to change my algorithm preference it in the initiator. Any hints?

I don't think they're merged yet, a rebased set was posted a few days ago.
https://marc.info/?l=target-devel&m=157226632013186&w=2

As for preferences, yeah I should probably do that.  Right now it just advertises all 3 if the OpenSSL libs seem to have working implementations.  Is hating our configuration code a good enough reason for that to be missing? ;-)

gonzoleeman:
> > Also, I'd like to test the newer authentication algorithms, but I don't know how! I thought LIO/targetcli-fb were already updated to support the newer algorithms, but I don't see the code! Plus I'm not sure how to change my algorithm preference it in the initiator. Any hints?
> 
> I don't think they're merged yet, a rebased set was posted a few days ago.
> https://marc.info/?l=target-devel&m=157226632013186&w=2
> 
> As for preferences, yeah I should probably do that. Right now it just advertises all 3 if the OpenSSL libs seem to have working implementations. Is hating our configuration code a good enough reason for that to be missing? ;-)

I will be glad to add that part.

I vote for merging in these coverity fixes.

gonzoleeman:
Note: I may tag a new version soon, so that these fixes can get downstream more easily.

cleech:
> Note: I may tag a new version soon, so that these fixes can get downstream more easily.

Not a bad idea.  I'm OK with rethinking the version scheme too, I think you mentioned maybe just calling the next one 2.1 and going to 2 level versions?

gonzoleeman:
Ok, if you're fine with "2.1", I'll do that.

As far as the coverity_scan branch, are you going to take care of regular merge and check? Also, can you send me info on how to get such a coverity check for my other project, open-isns? Thanks for doing this stuff!

PR #182, opened 2019-12-20
configuration support for CHAP algorithms

cleech:
Introduces support for preference lists in configuration files, and uses
that for the 'node.session.auth.chap_algs' setting.

This is also re-used for discovery authentication, rather than have two
different configurations.

It turns out that adding to the configuration file is way harder than adding new digest algorithms to the CHAP code itself was.  Having two sets of idbm.c to worry about depending on which commands are being run really doesn't help either.

I've tried breaking this every way I can think of, and I can't make it fall over anymore.
It seems far enough along to share.  Let me know if you spot anything that needs changing.

### DIFF
--- a/etc/iscsid.conf
+++ b/etc/iscsid.conf
@@ -57,6 +57,13 @@ node.leading_login = No
 # to CHAP. The default is None.
 #node.session.auth.authmethod = CHAP
 
+# To configure which CHAP algorithms to enable set
+# node.session.auth.chap_algs to a comma seperated list.
+# The algorithms should be listen with most prefered first.
+# Valid values are MD5, SHA1, SHA256, and SHA3-256.
+# The default is MD5.
+#node.session.auth.chap_algs = SHA3-256,SHA256,SHA1,MD5
+
 # To set a CHAP username and password for initiator
 # authentication by the target(s), uncomment the following lines:
 #node.session.auth.username = username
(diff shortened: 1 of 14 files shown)


gonzoleeman:
Chris, sorry I haven't tested this yet, but I will in the next week. Can you share the config settings for this, you used for testing?

cleech:
No problem, I wanted to get this out but then I've been traveling for the
past week.

I've tested the new node.session.auth.chap_algs setting both inherited from
iscsid.conf and set with isciadm.  It's value is a list, which should
select the advertised chap algorithms and their order. By default this
patch goes back to just MD5. The other thing that's a bit odd is this is
the only with setting reused for discovery, but it felt confusing to have a
separate setting for that.

Chris

cleech:
I'm going to go ahead and merge this.  If there are objections to having the default be MD5 only when there is no configuration setting, I'd like to hear them (my thought is that this way keeps compatibility with buggy targets on an upgrade while using the old config file).

gonzoleeman:
I have not tested it yet, but that’s my bad, and I trust your judgement. :)

I will still test it soon, and on the off chance I see anything I’ll just report it as a possible issue.

P.S. Thank you for doing the work!

> On Jan 8, 2020, at 1:54 PM, Chris Leech <notifications@github.com> wrote:
> 
> I'm going to go ahead and merge this. If there are objections to having the default be MD5 only when there is no configuration setting, I'd like to hear them (my thought is that this way keeps compatibility with buggy targets on an upgrade while using the old config file).
> 


-- 
Lee Duncan

"Man is the center of a circle of infinite circumference." -- John Lilly

PR #183, opened 2020-01-08
Fix bug with libopeniscsiusr.pc

phmccarty:
The library name is libopeniscsiusr, so the -l linker option should be `-lopeniscsiusr` instead of `-liscsiusr`.


cleech:
Thanks Patrick!

phmccarty:
You're welcome :)

PR #185, opened 2020-01-29
Fix compiler issues

gonzoleeman:
These changes are all aimed at fixing things that an updated gcc doesn't like, including compiling with "-fno-common". Also, 586 compilation errors are fixed, since we still support 32-bit Intel. Lastly, -Werror is added to "usr/Makefile" so that future errors are caught. Tested with gcc-9 on 64- and 32-bit Intel architectures.

There _should_ be no functional change, so only minor testing was done.

### DIFF
--- a/include/iscsi_err.h
+++ b/include/iscsi_err.h
@@ -4,7 +4,7 @@
 #ifndef _ISCSI_ERR_
 #define _ISCSI_ERR_
 
-enum {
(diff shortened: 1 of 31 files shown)


gonzoleeman:
This is a fairly straight-forward request, and should not change anything functionally, so I'm merging it.

PR #186, opened 2020-01-31
Avoid logout of iscsi boot session

knakamu:
OS is stuck after reboot command on iscsi boot environment. 
The root cause is that the systemd can't access the iscsi boot disk after iscsi service was stopped by
reboot command. So the systemd can't execute remaining jobs. As a result, OS is stuck.
This patch avoids logout of iscsi boot session.


gonzoleeman:
Thank you. Looks good to me.

PR #187, opened 2020-02-03
Check whether socket is opened successfully in find_vlan_dev func

ZhiqiangLiu26:
In find_vlan_dev func, socket should be checked before used.

Signed-off-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>


gonzoleeman:
There are a couple ioctl()s in there who's return value get ignored, as well. Thank you for fixing this.

PR #197, opened 2020-03-24
iscsi-iname: fix iscsi-iname -p access NULL pointer without given IQN…

wubo009:
iscsi-iname -p access NULL pointer without give IQN prefix.

# iscsi-iname -p
Segmentation fault

Signed-off-by: Wu Bo <wubo40@huawei.com>


gonzoleeman:
This was fixed by commit b32f59619c32. Apologies for not saying so earlier.

PR #207, opened 2020-05-05
Add iscsi-init.service

LorbusChris:
Per Fedora Packaging Guidelines [1], initial configuration of a service
should happen in a one-off init service in order to ensure idempotency,
and not in the %post directive of the RPM spec.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1493296

[1]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Initial_Service_Setup/

### DIFF
--- a/etc/systemd/iscsi-init.service
+++ b/etc/systemd/iscsi-init.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=One time configuration for iscsi.service
+ConditionPathExists=!/etc/iscsi/initiatorname.iscsi
+
+[Service]
+Type=oneshot
+RemainAfterExit=no
+ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`/usr/sbin/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
--- a/etc/systemd/iscsi.service
+++ b/etc/systemd/iscsi.service
@@ -2,9 +2,9 @@
 Description=Login and scanning of iSCSI devices
 Documentation=man:iscsiadm(8) man:iscsid(8)
 Before=remote-fs.target
-After=network.target network-online.target iscsid.service
-Requires=iscsid.socket
-ConditionPathExists=/etc/iscsi/initiatorname.iscsi
+After=network.target network-online.target
+After=iscsid.service iscsi-init.service
+Requires=iscsid.socket iscsi-init.service
 
 [Service]
 Type=oneshot


LorbusChris:
RPM Spec change: https://src.fedoraproject.org/rpms/iscsi-initiator-utils/pull-request/8

etc/systemd/iscsi.service
```diff
@@ -2,9 +2,9 @@
 Description=Login and scanning of iSCSI devices
 Documentation=man:iscsiadm(8) man:iscsid(8)
 Before=remote-fs.target
-After=network.target network-online.target iscsid.service
-Requires=iscsid.socket
-ConditionPathExists=/etc/iscsi/initiatorname.iscsi
+After=network.target network-online.target
```
gonzoleeman, line 5: Why did you remove the requirement to go after iscsid.service?

etc/systemd/iscsi-init.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=One time configuration for iscsi.service
+ConditionPathExists=!/etc/iscsi/initiatorname.iscsi
+
+[Service]
+Type=oneshot
+RemainAfterExit=no
+ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`/usr/sbin/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
```
gonzoleeman, line 8: Shouldn't you ensure that the initiator name matches the iBFT initiator name (from /sys/firmware/ibft/initiator/initiator-name)?

We use a distro-specific script to set up our initiator name, but perhaps it's time to make something like that script common?

etc/systemd/iscsi.service
LorbusChris: I did not remove it, I just moved it one line below into the new After= statement to make lines shorter and have a nice symmetry with the Requires= statement another line below.

etc/systemd/iscsi-init.service
LorbusChris: I must admit I didn't dive in too deep here. This command is the one currently used to create it on Fedora and RHEL (in %post of the specfile).
From Fedora: https://src.fedoraproject.org/rpms/iscsi-initiator-utils/blob/master/f/iscsi-initiator-utils.spec#_219

gonzoleeman: In SUSE, we have a script, created by Hannes, that augments creating the initiator name, called iscsi-gen-initiatorname, a shell script. It has a "-f/--force" option, but otherwise will not overwrite /etc/iscsi/initiatorname.iscsi file. If the file /sys/firmware/ibft/initiator/initiator-name exists it uses that as the initiator name and errors out if there is an existing initiator name that doesn't match (if not in "force" mode). Then it creates the initiatorname file with a proper comment about where the name came from (BIOS or not), how you should not change it, etc. When setting the initiator name (not from iBFT), it uses a template:

```
        ISSUEDATE="1996-04"
        INAME=$(/sbin/iscsi-iname -p iqn.$ISSUEDATE.de.suse:01)
        printf "InitiatorName=$INAME\n"  >>/etc/iscsi/initiatorname.iscsi
        chmod 0600 /etc/iscsi/initiatorname.iscsi
```

This is the reason, as far as I understand it, why this script is SUSE-specific. But it seems like the actual template could be distro-specific.

Perhaps, though, this can be treated as a separate issue from your pull request. Even so, do you think RedHat would be interested in using such a script? If so, I'd be glad to submit it. If not, that's fine too.

LorbusChris: Note that the init service will only run if the `/etc/iscsi/initiatorname.iscsi` file does not exist yet.
So if that file is generated in a different way in SUSE before the service is started, the init service wouldn't run at all.

I don't know why it's done that way in SUSE, and under what circumstances that might also make sense for Fedora/RHEL systems.

gonzoleeman: > Note that the init service will only run if the `/etc/iscsi/initiatorname.iscsi` file does not exist yet.
> So if that file is generated in a different way in SUSE before the service is started, the init service wouldn't run at all.
> 
> I don't know why it's done that way in SUSE, and under what circumstances that might also make sense for Fedora/RHEL systems.

It would only make a difference if they wanted to be able to customer the initiator name. Right now it appears they are using the open-iscsi default. Sometimes it's nice, for example in debugging, as well as in name recognition, to have the initiator IQN (name) include the company "name".

I am going to go ahead and approve this, but I'm pretty sure I'll be using a slightly different version of this in SLES. I'd love it if we could get all the distributions (SUSE, RedHat, Canonical, etc) to use exactly the same RPM/spec files and the same systemd service files, but I don't think that'll ever be the case.

PR #212, opened 2020-06-04
iscsi: fix fd leak

lixiaokeng:
iscsi-review/utils/fwparam_ibft/fwparam_ibft.c: line 466

Before return -1, the fd was opend and not closed which would lead
to leak fd. Fix that.
Signed-off-by: lixiaokeng<lixiaokeng@huawei.com>
Reported-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>


gonzoleeman:
Thank you.

PR #221, opened 2020-09-03
Check Invalid Session id for stop connection

gulams:
Description
-------------
If the initiator is rebooting then after the reboot, it will try to resync (recreate) the existing the connections by reading the sysfs. While initiator is doing this, i.e when the initiator tries to connect to the target but if the target service is not yet started, then the initiator connection will fail. The session id is also not yet assigned and it will be at its initial value 0xFFFFFFFF which is invalid. The session id is assigned a valid value only after a successful connection. Since the connection is failed, the initiator code will queue the connection for re-open. The connection state is still at ISCSI_CONN_STATE_XPT_WAIT as its very first login attemp after the reboot. 
Due to my Pull #206 request the code will invoke the stop connection to decrement the socket_fd reference count to properly close the connecion (details are in pull request #206). But since the session id is not valid, the stop connection will fail and the code will go ahead and queue the re-open without attempting the connect again. This is repeated till 120 seconds (stop connection failing and requeuing the reopen without invoking connect) and the connection will be shutdown resulting the storage unavailable.

Fix
---
We need to check the validity of the session id before calling the stop connection. If the session id is valid then only invoke the stop connection. Due to this, the code will go ahead and attempt the connet call. If the target service comes up anytime in 120 seconds, then the connect will be successful and we will get connected to the target.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -692,6 +692,7 @@ static void iscsi_login_eh(struct iscsi_conn *conn, struct queue_task *qtask,
 			   int err)
 {
 	struct iscsi_session *session = conn->session;
+	int stop_flag = 0;
 
 	log_debug(3, "iscsi_login_eh");
 	/*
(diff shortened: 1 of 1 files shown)


joejin00:
The issue could be reproduced by target and initiator on same server, and iscsid service start earlier then target, then iscsid can not login into local iscsi target.

We have verified the patch fix the issue.

joejin00 (approved):
stop_flag check condition is not correct.

gulams:
> 
> 
> stop_flag check condition is not correct.
I think the check condition is correct as the INVALID_SESSION_ID is defined as "(uint32_t)-1"

cleech:
Thank you for submitting this.  I think the change looks OK.

However, "Update initiator.c" is not a sufficient commit message.  Could you modify the commit in your tree to include the justification you provided in the PR comments?

cleech (changes requested):
commit description changes required

gulams:
> 
> 
> commit description changes required

Changed the commit description as requested

gonzoleeman:
It looks like you did what Chris asked.

gonzoleeman:
Thank you!

PR #229, opened 2020-11-02
iscsid: Poll timeout value to 1 minute for iscsid

gulams:
Description
    ===========
    This patch has the following two changes
    ------------------------------------------

    Change 1: Specify the poll timeout value to 1 minute as third parameter
    to the function iscsid_exec_req() when called from sync_session()

    Reason: Currently the poll timeout value sent sent to iscsid_response()
    from iscsid_exec_req() is "info->iscsid_req_tmo" which is -1 as set in
    "iscsi_sysfs_for_each_session()". When iscsid_response() receives this
    -1, it sets the timeout value to ISCSID_REQ_TIMEOUT (1000 ms) and also
    sets a local variable "poll_wait" to 1. There is a while loop below this
    which checks the value of "poll_wait". If "poll_wait" is set to 1, then
    it calls "continue". For sessions which are giving continuous connection
    errors (like target service stopped OR target node is shutdown OR any
    other continuous connection errors etc ...), this results in an
    indefinite while loop, as nothing will be written to the poll fd,
    resulting in further sessions not getting synced (or recovered) when the
    iscsid is restarted due to any reason (either manual restart of iscsid
    or rpm install). Poll timeout of 1 minute seems to be a reasonable value
    for the slow connections

    Change 2: Change the return error code from the function
    iscsid_response() to ISCSI_ERR_SESSION_NOT_CONNECTED when the poll times
    out and returns error 0

    Reason: Currently the iscsid_response() function returns
    ISCSI_ERR_ISCSID_NOTCONN error code in case if the poll times out and
    the poll_wait variable is 0 (i.e poll timeout sent to
    iscsid_response() was -1). Returning this error code doesn't
    seem to be correct for the following two reasons:
        a. ISCSI_ERR_ISCSID_NOTCONN should be returned only when we
           are not able to connect to iscsid but we are in iscsid_response()
           function indicating that we were able to  connect to iscsid
           successfully
        b. When ISCSI_ERR_ISCSID_NOTCONN is returned, the
           sync_session() will retry the request till 30 retries are reached. This
           causes an overlap of multiple "iscsi login task" in
           kernel and the kernel will return an error to the
           user-space indicating "Login/Text in progress. Cannot start new task.".
           This is repeated continuously which results in session not getting
           recovered even after it comes back up. Also at one point of time we
           observed that the kernel panics in "iscsi_sw_tcp_conn_set_param()" while
           trying to set the param ISCSI_PARAM_DATADGST_EN. We have a kernel fix
           for this panic which is in review for upstream

### DIFF
--- a/usr/iscsi_sysfs.c
+++ b/usr/iscsi_sysfs.c
@@ -1426,7 +1426,7 @@ int iscsi_sysfs_for_each_session(void *data, int *nr_found,
 	if (!info)
 		return ISCSI_ERR_NOMEM;
 
-	info->iscsid_req_tmo = -1;
+	info->iscsid_req_tmo = ISCSID_RESP_POLL_TIMEOUT;
 	n = scandir(ISCSI_SESSION_DIR, &namelist, trans_filter,
 		    alphasort);
 	if (n <= 0)
--- a/usr/iscsi_sysfs.h
+++ b/usr/iscsi_sysfs.h
@@ -34,6 +34,7 @@ struct iscsi_auth_config;
 struct flashnode_rec;
 
 #define SCSI_MAX_STATE_VALUE 32
+#define ISCSID_RESP_POLL_TIMEOUT 60000
 
 extern void free_transports(void);
 extern char *iscsi_sysfs_get_iscsi_kernel_version(void);
--- a/usr/iscsid_req.c
+++ b/usr/iscsid_req.c
@@ -156,7 +156,7 @@ int iscsid_response(int fd, iscsiadm_cmd_e cmd, iscsiadm_rsp_t *rsp,
 		if (!err) {
 			if (poll_wait)
 				continue;
-			return ISCSI_ERR_ISCSID_NOTCONN;
+			return ISCSI_ERR_SESSION_NOT_CONNECTED;
 		} else if (err < 0) {
 			if (errno == EINTR)
 				continue;


gonzoleeman:
LOL. I don't think iscsid_req_tmo is ever set to anything but -1. Kind of silly. And the code in iscsid_response() if you pass in "-1" is kind of dumb too IMHO. Retry forever? [I vehemently dislike retry-forever.] If so, why time out every 1 second then go do it again? Why not just do a poll() without a timeout? Strange. I fear there are reasons I don't grok.

I would like to see this whole mess cleaned up, but I won't discount your patch because of that. :)

I don't see anything wrong with the idea of your fix, but I'd like to change the assignment in iscsi_sysfs_for_each_session() to "60 seconds" instead of overriding it in sync_session(). And please make it a define (like ISCSID_REQ_TIMEOUT, but with a better name).

Lastly, a nit on your verbiage in your commit: You say:

"Specify the poll timeout value to 1 minute as third parameter
    to the function iscsid_exec_req() when called from sync_session()"

But it's actually the fourth parameter I believe.

gulams:
Thanks Lee for the review. 

 

Can you please see me inline comments?

 

Also, this patch has one more change (I mentioned it as a “Change 2”). Can you please give your comments on that also?

 

Regards,

Gulam Mohamed.

gonzoleeman:
You asked me to comment on Change 2. I didn't say anything because that looked fine to me. ISCSI_ERR_ISCSID_NOTCONN is defined as "Could not connect to iscsid", which clearly isn't what's happening here.

gulams:
Did a force push after completing the suggested changes and testing. Can you please look into this?

 

Regards,

Gulam Mohamed.

gonzoleeman:
You asked me to comment on Change 2. I didn't say anything because that looked fine to me. ISCSI_ERR_ISCSID_NOTCONN is defined as "Could not connect to iscsid", which clearly isn't what's happening here.

gonzoleeman:
Thank you!

zeha:
Debian bug 980085 (https://bugs.debian.org/980085) claims 2.1.3 broke iscsi login on some systems (from 2.1.2). This PR might be related. Also there's someone on stackexchange seeing a similar problem, referencing this PR.

https://unix.stackexchange.com/questions/626162/iscsi-login-issue-32-target-not-likely-connected 

@gulams could you maybe check these out? Thanks!

zeha:
FTR, I've locally reverted this patch, and can confirm the timeout behaviour has changed. In 2.1.2, one had a little more time for the network to become ready ("link up") - in 2.1.3, if the link is not ready yet, it just immediately fails.

gulams:
> 
> 
> FTR, I've locally reverted this patch, and can confirm the timeout behaviour has changed. In 2.1.2, one had a little more time for the network to become ready ("link up") - in 2.1.3, if the link is not ready yet, it just immediately fails.

I am looking at this and will update soon.

gulams:
> 
> 
> The change in usr/iscsid_req.c 159 breaks booting from iSCSI using open-iscsi 2.1: iscsistart:
> 
> ```
> initiator reported error (32 - target likely not connected)
> ```
> 
> Cf. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980085

I tried reproducing this based on the instructions provided in this link https://github.com/open-iscsi/open-iscsi/issues/241
by Lee Duncan, but couldn't. I noticed that he has provided another patch with a change of return value check in login_session() along with a debug print.

I will continue my efforts to reproduce this and understand.

PR #230, opened 2020-11-04
Fix memory leak in sysfs_get_str

wenchao-hao:
strlen(sysfs_value) return 0 do not mean sysfs_value
is NULL and this memory should be freed


gonzoleeman:
Thank you for your patch. This seems unlikely (having a zero-length sysfs attribute), but this can't hurt.

PR #231, opened 2020-11-11
iscsid: fix logging level when starting and shutting down daemon

yoshkosh:
Some users monitor syslog by logging level. Inappropriate logging level breaks their monitoring because of false alert.


gonzoleeman:
Thank you!

PR #233, opened 2020-11-23
iscsiadm: Optimize the the verification of mode paramters

wenchao-hao:
The mode parameters should be check earlier when recognize the
mode, rather than mutiple calls after all commandline parameters
are parsed.

Signed-off-by: Wu Bo <wubo40@huawei.com>
Signed-off-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>

### DIFF
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -101,6 +101,22 @@ enum _print_node_tree_mode {
 	_PRINT_MODE_NODE,
 };
 
+struct verify_mode_t {
+	char *mode;
+	char *allowed;
+	int skip_m;
+};
+
+static const struct verify_mode_t mode_paras[] = {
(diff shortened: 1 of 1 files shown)


usr/iscsiadm.c
```diff
@@ -101,6 +101,22 @@ enum _print_node_tree_mode {
 	_PRINT_MODE_NODE,
 };
 
+struct verify_mode_t {
+	char *mode;
+	char *allowed;
+	int skip_m;
+};
+
+static const struct verify_mode_t mode_paras[] = {
```
gonzoleeman, line 110: "mode_paras" isn't consistent with the other variables and functions, which call them "mode_params". But this is a nit. You do not have to fix this, but keep it in mind for next time.

gonzoleeman:
I made one review comment, but it's a nit. I'll wait another day or two, for @cleech to get a chance to comment, if he wishes.

PR #234, opened 2020-12-12
fix serverl issues reported by Coverity

wenchao-hao:
Recently, we use Coverity to analysis the open-iscsi package.
Several issues should be resolved to make Coverity happy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,11 @@
 *.o
 cscope.files
 cscope.out
+cscope.in.out
+cscope.po.out
+GPATH
+GRTAGS
+GTAGS
 TAGS
 tags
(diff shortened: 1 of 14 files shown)


gonzoleeman (commented):
The wording on this commit is not good. Perhaps you can have somebody help you check the syntax and spelling before submitting next time?

gonzoleeman:
Some of these commits are a bit unneeded, such as freeing memory before exiting. But all of them look clean, and some of them look helpful.

wenchao-hao:
> The wording on this commit is not good. Perhaps you can have somebody help you check the syntax and spelling before submitting next time?

Yes, I would.

PR #236, opened 2020-12-19
Fix two issues in iscsid 

wenchao-hao:
This PR fixes two issues in iscsid
1. change iscsid writing pidfile from child process to parent process to avoid systemd report error with fork mode;
2. add resource clean before process exit

### DIFF
--- a/usr/iscsid.c
+++ b/usr/iscsid.c
@@ -451,6 +451,7 @@ int main(int argc, char *argv[])
 
 	if ((mgmt_ipc_fd = mgmt_ipc_listen()) < 0) {
 		log_close(log_pid);
+		idbm_terminate();
 		exit(ISCSI_ERR);
 	}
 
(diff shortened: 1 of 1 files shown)


gonzoleeman:
Thank you for your submission!

Your first commit looks fine, though perhaps a little pedantic, as closing file descriptors before an exit() is perhaps good practice but does nothing, as exit() takes care of that. Never the less, such cleanups don't hurt, and I'll be glad to merge it.

But your second commit has an issue. If I supply "-n" (no pid file), your code causes iscsid to core dump. This is because "fd" never gets set if "-n" (no pid file) is supplied, but you assume "fd" is valid. The original code did not have this issue.

Please fix this and resubmit. Note: I just added mainline code to iscsid to make sure nobody supplies both "-n" and "-p", since the original code _did_ have an issue with that.

PR #237, opened 2020-12-27
RFC: Regression in 2.1.3 causes `iscsistart` to fail to login to sessions

mwberry:
In [this commit](https://github.com/open-iscsi/open-iscsi/commit/9258c8eae046d98511d92912983778ca57ba201f), which landed in 2.1.3, the method `iscsid_response` in `iscsid_req.c` changed a return code from `ISCSI_ERR_ISCSID_NOTCONN` to `ISCSI_ERR_SESSION_NOT_CONNECTED`. There are two reasons stated in the commit message and they seem legitimate. However, it seems `iscsistart` relied on the old behavior in `login_session` and it was not updated to match. This manifested in an annoying fashion. During a UEFI network boot, which invoked `iscsistart`, the initiator would open a TCP connection to the target and then a second later it would close that TCP connection without sending any data on it. The initiator would then complain with the error `initiator reported error (32 - target likely not connected)`. I believe this is because the underlying TCP transport is set to non-blocking, so the first `ep_connect` returns `EINPROGRESS`. Under normal circumstances it would eventually get resumed. However since the errant `ISCSI_ERR_SESSION_NOT_CONNECTED` error is returned out of `login_session` / `setup_session` and the child thread is killed before any progress can happen. It just so happened that on a single-core VM where I was testing, the race was reliably won and I was unable to reproduce the issue. But on real multi-core hardware I'd hit the error case every time.

I corrected that oversight and grep'd for remaining usages of the `ISCSI_ERR_ISCSID_NOTCONN`. The remaining uses all come from `ipc_connect`, which seems appropriate. Only `sync_session` seems to catch this error. I'm not familiar with the code however so someone should double check that.

### DIFF
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -246,7 +246,7 @@ static int login_session(struct node_rec *rec)
 		rc = iscsid_exec_req(&req, &rsp, 0, ISCSID_REQ_TIMEOUT);
 		if (rc == 0) {
 			return rc;
-		} else if (rc == ISCSI_ERR_ISCSID_NOTCONN) {
+		} else if (rc == ISCSI_ERR_SESSION_NOT_CONNECTED) {
 			ts.tv_sec = msec / 1000;
 			ts.tv_nsec = (msec % 1000) * 1000000L;
 


mwberry:
Unfortunately, further testing actually indicates this fix will not work. It seems that retrying at `login_session` is too high-level. On the second attempt it generates the error `15 - session exists`, suggesting something lower-level needs to account for the time it takes to open the TCP connection. 

The minimal diff which fixes the issue for me is not a proper fix, but a workaround:

`iscsistart.c`:
```
246c246
<               rc = iscsid_exec_req(&req, &rsp, 0, ISCSID_REQ_TIMEOUT);
---
>               rc = iscsid_exec_req(&req, &rsp, 0, -1);
```

This causes `iscsid_response` to continue to `poll` locally. I don't think this is really appropriate. I'm still reading through the code, but it looks like there's some machinery in `initiator.c` and elsewhere around establishing the connection in an asynchronous manner. For example the tail end of `__session_login_task` looks like it will queue up additional callbacks if it gets `EINPROGRESS` however I had my log level set to 6 and never saw this message.


```
        if (iscsi_conn_connect(conn, qtask)) {
                log_debug(4, "Initial connect failed. Waiting %u seconds "
                          "before trying to reconnect.",
                          ISCSI_CONN_ERR_REOPEN_DELAY);
                queue_delayed_reopen(qtask, ISCSI_CONN_ERR_REOPEN_DELAY);
        }

        return ISCSI_SUCCESS;
```

gonzoleeman:
Can you try this diff?

```
diff --git a/usr/iscsistart.c b/usr/iscsistart.c
index 73991b37cb8b..813fa743937d 100644
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -246,7 +246,8 @@ static int login_session(struct node_rec *rec)
                rc = iscsid_exec_req(&req, &rsp, 0, ISCSID_REQ_TIMEOUT);
                if (rc == 0) {
                        return rc;
-               } else if (rc == ISCSI_ERR_ISCSID_NOTCONN) {
+               } else if ((rc == ISCSI_ERR_ISCSID_NOTCONN) ||
+                          (rc == ISCSI_ERR_SESSION_NOT_CONNECTED)) {
                        ts.tv_sec = msec / 1000;
                        ts.tv_nsec = (msec % 1000) * 1000000L;

```

In other words, catch both ISCSI_ERR_ISCSID_NOTCONN and ISCSI_ERR_SESSION_NOT_CONNECTED.

gonzoleeman:
Closing this, since the issue is resolved.

PR #238, opened 2020-12-28
Add etc/systemd/iscsi-init.service to SYSTEMDFILES Makefile variable

jnschaeffer:
iscsi-init.service is not included in the `SYSTEMDFILES` Makefile variable, causing it to be skipped when the `install_systemd` rule is executed. This causes the `iscsi` systemd service to fail on a fresh install, as it requires the `iscsi-init` service.

This PR adds `etc/systemd/iscsi-init.service` to the list of files in `SYSTEMDFILES` to correct this issue.


gonzoleeman:
Thank you for your fix!

PR #246, opened 2021-01-25
iscsid: Add NO_SYSTEMD to CFLAGS

B4dM4n:
When building with `NO_SYSTEMD=1` set the `-DNO_SYSTEMD` `CFLAG`, otherwise the `iscsid` build tries to include the systemd header and aborts:

```
iscsid.c:38:10: fatal error: systemd/sd-daemon.h: No such file or directory
```


gonzoleeman:
Thank you.

PR #247, opened 2021-02-01
Wants=network-online.target in iscsi.service

matwey:
Setting After= without Requires= or Wants= is not enough to pull
network-online.target as a dependency.

Currently, if some other service requires network-online.target
then iscsi.service is started after network-online.target

If no any other services require network-online.target then
iscsi.service may be started before networking is working.
The latter leads to the following issues:

    iscsiadm: Could not login to [iface: default, target: XXX, portal: XXX,3260].
    iscsiadm: initiator reported error (4 - encountered connection failure)

Reference: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/


gonzoleeman:
Thanks you.

PR #250, opened 2021-02-15
Avoid hardcoding pkg-config to fix cross build

zeha:
Debian Bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=982307

Patch from @helmutg.


gonzoleeman:
Looks good. Thank you!

PR #257, opened 2021-04-05
iscsid: Log error message when auth debug status is set

gulams:
Log the correct error message in /var/log/messages when the debug status is set when there is an authentication error.The change is made in check_security_stage_status() function. This issue was identified when one of the customer was trying to login to the iscsi target after configuring the mutual CHAP authentication. They set the same password for both side authentication. This is not allowed as per the code as it was returning AUTH_DBG_STATUS_PASSWD_IDENTICAL. The error logged in /var/log/messages was not obvious. This change will help to understand the correct error message so that required action can be taken.

### DIFF
--- a/usr/login.c
+++ b/usr/login.c
@@ -624,8 +624,8 @@ check_security_stage_status(iscsi_session_t *session,
 	case AUTH_STATUS_ERROR:
 	case AUTH_STATUS_FAIL:
 	default:
-		acl_get_dbg_status(auth_client, &debug_status);
-		if (debug_status !=   AUTH_DBG_STATUS_NOT_SET)
+		if (acl_get_dbg_status(auth_client, &debug_status) == 
+			AUTH_STATUS_NO_ERROR)
 			log_error("Login authentication failed "
 				       "with target %s, %s",
 				       session->target_name,
--- a/usr/login.c
+++ b/usr/login.c
@@ -624,8 +624,8 @@ check_security_stage_status(iscsi_session_t *session,
 	case AUTH_STATUS_ERROR:
 	case AUTH_STATUS_FAIL:
 	default:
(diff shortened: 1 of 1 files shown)


usr/login.c
```diff
@@ -624,8 +624,8 @@ check_security_stage_status(iscsi_session_t *session,
 	case AUTH_STATUS_ERROR:
 	case AUTH_STATUS_FAIL:
 	default:
-		if (acl_get_dbg_status(auth_client, &debug_status) !=
-		    AUTH_STATUS_NO_ERROR)
+		acl_get_dbg_status(auth_client, &debug_status);
+		if (debug_status !=   AUTH_DBG_STATUS_NOT_SET)
```
gonzoleeman, line 628: The original code seems backwards, but otherwise correct, or am I missing something?

It seems like it should just need the "!=" changed to an "==" when looking for AUTH_STATUS_NO_ERROR.

So to even get here, we know there is an auth error or failure. So we try to get the debug status. If we succeed in getting it, we should print out the result, even if it is "debug status not set". If we can't get it, then we just print a generic failure message.

But since the sense of the comparison is backwards, the result was that if we could get the debug status we wanted, then we printed the generic message. So the bug causes helpful information to be lost.

With your suggested fix, you do not check the result from acl_get_dbg_status(), just assuming it works.

gulams: Hi Lee,

     I did the suggested change and force pushed it. Can you please review? If  you want me to close this pull request and open a new one with the new changes, please let me know.

Regards,
Gulam Mohamed.

gonzoleeman:
Yes, this request has become messy from the forced-push. I'll close this. Please submit a new pull request and I'll be glad to merge it.

PR #264, opened 2021-05-11
Set default 'startup' to 'onboot' for FW nodes

gonzoleeman:
When "iscsiadm -m discovery -t fw" is ran, nodes are
created/updated in the node DB for each firmware target
found, but the 'startup' mode values are left as 'manual',
when in fact firmware nodes are treated as if this
value is 'onboot'. So to keep the node DB in sync with
behavior, set these to 'onboot' by default. This
should *not* effect any other functionality.


gonzoleeman:
Since there was no comment/objection, merged.

PR #265, opened 2021-06-16
iscsistart: fix null pointer deref before exit

cleech:
Fixes regression caused by "open-iscsi: Clean user_param list when
process exit"  Which is a shame, as not freeing a memory at process exit
doesn't really hurt anything.

Same change as "Fix iscsiadm segfault when exiting" applied to iscsiadm.

Fixes: b532ad67d495d42026165a26515c645995d23f18


gonzoleeman:
Thanks Chris! I thought I'd fixed that. :(

PR #269, opened 2021-07-28
Add prctl(PR_SET_IO_FLUSHER) to avoid mem deadlock

gonzoleeman:
Use the new prctl(PR_SET_IO_FLUSHER, 1) system call call
o ensure that memory allocation, when interacting with our
ernel components, doesn't cause a deadlock, but only if
it is available.


gonzoleeman:
@cleech ??? Does this look ok to you? It compiles on pre- and post-5.6 kernels/systems.

gonzoleeman:
Mike's approach used, instead.

PR #273, opened 2021-09-06
[old] iscsistart: Fix iscsistart timeout during login session

jkhasdev-ora:
When iscsistart places a login request during boot, if it gets timeout
and response receive after it. It triggers "connection1:0: detected conn
error (1011)", since in userspace nobody is listening.

On OL6, it has been observed where iscsid is not part of initramfs.

During boot, iscsistart forks and child process used as placing a request whereas
parent wait in event loop.

In the child, login_seesion places login request, and request gets
timed-out then parent also exit. If it receive response for timed-out
request and during that time no parent is listening to it then it will
error as 1011.

This patch is to increase tmo for iscsistart login request to x6. This
patch has been tested some intesive workload and helps resolve the
issue.

Signed-off-by: Jitendra Khasdev <jitendra.khasdev@oracle.com>

### DIFF
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -254,7 +254,7 @@ static int login_session(struct node_rec *rec)
 	 * login.
 	 */
 	for (msec = 50; msec <= 15000; msec <<= 1) {
-		int tmo = ISCSID_REQ_TIMEOUT * 10;
+		int tmo = ISCSID_REQ_TIMEOUT * 60;
 
 		rc = iscsid_exec_req(&req, &rsp, 0, tmo);
 		if (rc == 0) {


gonzoleeman:
I do not use iscsistart, nor does the distro I work on (SUSE), since using iscsid/iscsiadm is superior. The iscsistart command was, I believe, an attempt to handle just the iscsi login without needing iscsid, but it kind of sucks at that, as this bug demonstrates. The proper fix is to use iscsid/iscsiadm IMHO. But I realize that end users do not have that option if the distro has designed their implementation to use iscsistart.

For this issue and iscsistart, I'm not sure I agree in a blanket change of the timeout to 60 seconds from 10 seconds. That can directly impact how long it takes to time out if the target is in fact missing, i.e. that means a one-minute hang instead of a 10-second hang in that case. Perhaps this would be better handled by adding a timeout option to iscsistart (although that will also require your distro to do a fix).

jkhasdev-ora:
Lee, if I correctly understood then you are suggesting a "timeout" parameter to iscsistart and make changes into our distro to support it.

gonzoleeman:
> Lee, if I correctly understood then you are suggesting a "timeout" parameter to iscsistart and make changes into our distro to support it.

Correct. Under the principal of least surprise: this way, other current users of iscsistart will not see any changes unless they want to (by passing in a new timeout value).

Another option would be a new iscsid.conf attribute for this iscsistart timeout, that would override the default, but that seems like more work.

jkhasdev-ora:
sure Lee, I checked in initramfs of OL6, which does not have iscsid.conf to read, so I went ahead with the safest option by adding a new argument to iscsistart.

jkhasdev-ora:
Let me know if I can close this request because I have initiated a fresh PR for adding a new argument.

gonzoleeman:
Closing.

PR #275, opened 2021-09-13
iscsistart: Fix iscsistart timeout during login session

jkhasdev-ora:
When iscsistart places a login request during boot, if it gets timeout
and response receive after it. It triggers "connection1:0: detected conn
error (1011)", since in userspace nobody is listening.

On OL6, it has been observed where iscsid is not part of initramfs.

During boot, iscsistart forks and child process used as placing a request whereas
parent wait in event loop.

In the child, login_seesion places login request, and request gets
timed-out then parent also exit. If it receive response for timed-out
request and during that time no parent is listening to it then it will
error as 1011.

This patch adds an additional option to increase tmo for login request
timeout. This patch has been tested some intesive workload by increase it as 6x
and helps resolve the issue.

Signed-off-by: Jitendra Khasdev <jitendra.khasdev@oracle.com>

### DIFF
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -65,7 +65,7 @@ static char program_name[] = "iscsistart";
 /* used by initiator */
 extern struct iscsi_ipc *ipc;
 
-static int login_req_timeout = 10;
+static int login_req_timeout = 10000;
 
 static struct option const long_options[] = {
 	{"initiatorname", required_argument, NULL, 'i'},
@@ -258,7 +258,7 @@ static int login_session(struct node_rec *rec)
 	 * login.
 	 */
 	for (msec = 50; msec <= 15000; msec <<= 1) {
-		int tmo = ISCSID_REQ_TIMEOUT * login_req_timeout;
+		int tmo = login_req_timeout;
 
 		rc = iscsid_exec_req(&req, &rsp, 0, tmo);
 		if (rc == 0) {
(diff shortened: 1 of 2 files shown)


gonzoleeman:
I closed the other ("[OLD]") pull request, and now this one seem  broken. When I try to look at the diffs, it says HTML 404 error page not found?

usr/iscsistart.c
```diff
@@ -254,7 +258,7 @@ static int login_session(struct node_rec *rec)
 	 * login.
 	 */
 	for (msec = 50; msec <= 15000; msec <<= 1) {
-		int tmo = ISCSID_REQ_TIMEOUT * 10;
+		int tmo = ISCSID_REQ_TIMEOUT * login_req_timeout;
```
gonzoleeman, line 261: I don't like multiplying ISCSID_REQ_TIMEOUT times the login_req_timeout. It just so happens that currently ISCSID_REQ_TIMEOUT is one second, but what if it changes to 5 seconds? Then your iscsistart timout would multiply times 5! Can you just login_req_timeout not depend on ISCSID_REQ_TIMEOUT?

gonzoleeman (changes requested):
The issistart man page also needs to be updated. Please add that.

gonzoleeman:
One last small change requested. Can you please update your branch so I can pull your request?

mikechristie:
Hey Lee, jkhasdev-ora is no longer at oracle and is free from iscsi :)

Me on the other hand.... is hopefully temporarily helping out :)

This patch is wrong. I'm going to close this and will open a new PR.

PR #276, opened 2021-09-18
Fix recent commit gcc error

gonzoleeman:


gonzoleeman:
This is an error. I will redo.

Issue #282, opened 2021-10-26
The open-iscsi iscsi-init.service service cannot work if run too early


gonzoleeman:
In openSUSE Tumbleweed, if one removes the /etc/iscsi/iniatorname.iscsi file (which contains the system's IQN or iscsi name), and then reboots, the file will not be recreated, despite the fact that this is the purpose of the iscsi-init.service.

The reason is that this service is run quite early, and the root filesystem can still be read-only. In this case, the service tries to run but fails, so the other services that depend on it cannot run (such as iscsid.service, and hence iscsi.service).

The iscsi-init.service service files needs something like "After=root.mount" to work correctly. In testing on Tumbleweed on a VM, this fixed the issue.

gonzoleeman:
This was handled by pull request #285, so marking as closed.

gonzoleeman:
Fixed with commit:

> e6e53d6e60bf Fix iscsi-init so that it runs when root writable

PR #285, opened 2021-11-10
Fix iscsi-init so that it runs when root writable

gonzoleeman:
A recent commit, 432bbf979ee6 ("Remove dependences
from iscsi-init.service") removed DefaultDependencies from
iscsi-init.service, but that now means it can run so
early that the root disc is not yet writable, rendering
it useless, since it can't create the initiatorname.iscsi
file. This change tells it to wait until root is writable
to run.


gonzoleeman:
Addresses Issue #282

PR #287, opened 2021-11-19
iscsiadm: Call log_init() first to fix a segmentation fault

wenchao-hao:
log_init() should be called before log_error() been called


gonzoleeman:
Looks good. Thanks you.

PR #288, opened 2021-11-23
ipc: Introduce iscsid_uniq_id to avoid residual command being executed

wenchao-hao:
systemd would listen on ISCSIADM_ABSTRACT_NAMESPACE if iscsid.socket is started.
In this case, if iscsid exit with some iscsiadm_req_t requests unhandled, these
requests would not been cleared. When iscsid starts again, it would get these
residual requests and execute them.

It is easy to recurrent above phenomenon by following steps:

systemctl start iscsid.socket
systemctl start iscsid.service
killall -19 iscsid
systemctl restart iscsid

This PR fix above phenomenon by introducing check mechanism. Each iscsid has
a unique id, this unique id would be used to check if the requests should be
handled.

### DIFF
--- a/usr/iscsid.c
+++ b/usr/iscsid.c
@@ -448,6 +448,9 @@ int main(int argc, char *argv[])
 	if (log_pid < 0)
 		exit(ISCSI_ERR);
 
+	if (mgmt_init_iscsid_uniq_id())
+		exit(ISCSI_ERR);
+
 	/* do not allow ctrl-c for now... */
 	sa_new.sa_handler = catch_signal;
 	sigemptyset(&sa_new.sa_mask);
--- a/usr/iscsid_req.c
+++ b/usr/iscsid_req.c
@@ -38,6 +38,9 @@
 #include "iscsid_req.h"
 #include "uip_mgmt_ipc.h"
 
+int iscsid_response(int fd, iscsiadm_cmd_e cmd, iscsiadm_rsp_t *rsp,
+		    int timeout);
+
 static void iscsid_startup(void)
 {
 	char *startup_cmd;
@@ -118,13 +121,50 @@ static int iscsid_connect(int *fd, int start_iscsid)
 	return ipc_connect(fd, iscsid_namespace, start_iscsid);
 }
 
+static int iscsid_get_iscsid_uniq_id(int start_iscsid, uint32_t *iscsid_uniq_id)
+{
+	int err = 0;
+	int fd = 0;
+	iscsiadm_req_t req;
+	iscsiadm_rsp_t rsp;
+
+	err = iscsid_connect(&fd, start_iscsid);
+	if (err)
+		return err;
+
+	memset(&req, 0, sizeof(req));
+	req.command = MGMT_IPC_GET_ISCSID_UNIQ_ID;
+
+	if ((err = write(fd, &req, sizeof(req))) != sizeof(req)) {
+		log_error("write cmd MGMT_IPC_GET_ISCSID_UNIQ_ID failed: %s",
+			strerror(errno));
+		close(fd);
+		return ISCSI_ERR_ISCSID_COMM_ERR;
+	}
+
+	err = iscsid_response(fd, MGMT_IPC_GET_ISCSID_UNIQ_ID, &rsp, ISCSID_REQ_TIMEOUT);
+	if (err)
+		return err;
+
+	*iscsid_uniq_id = rsp.u.iscsid_uniq_id;
+
+	return 0;
+}
+
 int iscsid_request(int *fd, iscsiadm_req_t *req, int start_iscsid)
 {
 	int err;
+	uint32_t iscsid_uniq_id = 0;
+
+	err = iscsid_get_iscsid_uniq_id(start_iscsid, &iscsid_uniq_id);
+	if (err)
+		return err;
 
 	err = iscsid_connect(fd, start_iscsid);
 	if (err)
 		return err;
+
+	req->iscsid_uniq_id = iscsid_uniq_id;
 
 	if ((err = write(*fd, req, sizeof(*req))) != sizeof(*req)) {
 		log_error("got write error (%d/%d) on cmd %d, daemon died?",
--- a/usr/mgmt_ipc.c
+++ b/usr/mgmt_ipc.c
@@ -27,6 +27,7 @@
 #include <pwd.h>
 #include <sys/un.h>
 #include <string.h>
+#include <openssl/rand.h>
 
 #include "iscsid.h"
 #include "idbm.h"
@@ -43,6 +44,23 @@
 #define PEERUSER_MAX	64
 #define EXTMSG_MAX	(64 * 1024)
 #define SD_SOCKET_FDS_START 3
+
+static uint32_t iscsid_uniq_id;
+
+int mgmt_init_iscsid_uniq_id(void)
+{
+	int err = 0;
+
+	err = RAND_bytes((unsigned char *)&iscsid_uniq_id, 4);
+	if (err != 1) {
+		log_error("failed to generate iscsid_uniq_id");
+		return -1;
+	}
+
+	log_debug(3, "iscsid_uniq_id is 0x%x\n", iscsid_uniq_id);
+
+	return 0;
+}
 
 int
 mgmt_ipc_listen(void)
@@ -353,6 +371,18 @@ mgmt_ipc_notify_del_portal(queue_task_t *qtask)
 }
 
 static int
+mgmt_ipc_get_iscsid_uniq_id(queue_task_t *qtask)
+{
+	qtask->rsp.command = MGMT_IPC_GET_ISCSID_UNIQ_ID;
+	qtask->rsp.err = ISCSI_SUCCESS;
+	qtask->rsp.u.iscsid_uniq_id = iscsid_uniq_id;
+
+	mgmt_ipc_write_rsp(qtask, ISCSI_SUCCESS);
+
+	return ISCSI_SUCCESS;
+}
+
+static int
 mgmt_peeruser(int sock, char *user)
 {
 	/* Linux style: use getsockopt(SO_PEERCRED) */
@@ -482,6 +512,7 @@ static mgmt_ipc_fn_t *	mgmt_ipc_functions[__MGMT_IPC_MAX_COMMAND] = {
 [MGMT_IPC_NOTIFY_DEL_NODE]	= mgmt_ipc_notify_del_node,
 [MGMT_IPC_NOTIFY_ADD_PORTAL]	= mgmt_ipc_notify_add_portal,
 [MGMT_IPC_NOTIFY_DEL_PORTAL]	= mgmt_ipc_notify_del_portal,
+[MGMT_IPC_GET_ISCSID_UNIQ_ID]	= mgmt_ipc_get_iscsid_uniq_id,
 };
 
 void mgmt_ipc_handle(int accept_fd)
@@ -516,6 +547,19 @@ void mgmt_ipc_handle(int accept_fd)
 
 	command = qtask->req.command;
 	qtask->rsp.command = command;
+
+	/*
+	 * if iscsid_uniq_id do not match nor command type MGMT_IPC_GET_ISCSID_UNIQ_ID, then
+	 * this request is not send to current iscsid(may be a previous killed iscsid)
+	 * do not handle it
+	 */
+	if ((qtask->req.iscsid_uniq_id != iscsid_uniq_id) &&
+	    (command != MGMT_IPC_GET_ISCSID_UNIQ_ID)) {
+		log_error("mismatch iscsid_uniq_id: 0x%x desired while 0x%x provided",
+			iscsid_uniq_id, qtask->req.iscsid_uniq_id);
+		mgmt_ipc_destroy_queue_task(qtask);
+		return;
+	}
 
 	if (command > 0 &&
 	    command < __MGMT_IPC_MAX_COMMAND)
--- a/usr/mgmt_ipc.h
+++ b/usr/mgmt_ipc.h
@@ -46,6 +46,7 @@ typedef enum iscsiadm_cmd {
 	MGMT_IPC_NOTIFY_DEL_NODE	= 17,
 	MGMT_IPC_NOTIFY_ADD_PORTAL	= 18,
 	MGMT_IPC_NOTIFY_DEL_PORTAL	= 19,
+	MGMT_IPC_GET_ISCSID_UNIQ_ID	= 20,
 
 	__MGMT_IPC_MAX_COMMAND
 } iscsiadm_cmd_e;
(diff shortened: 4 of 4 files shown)


wenchao-hao:
> Why do you feel it's an error for the newly-started iscsid to finish requests from a previous incarnation? What error does this cause?

In automation scenario, scripts are used to execute the isciadm command
such as login or logout. iscsiadm returns error if the request is not processed.
In this case, the upper layer determines the execution of the iscsiadm fails. 
If newly-started iscsid process these residual commands, the final result
is inconsistent with what upper layer got.

If MGMT_IPC_IMMEDIATE_STOP remains, newly-started iscsid would exits
immediately after being started.

If login or logout commands remains, the actually session status is different
with what upper layer got. Which might lead upper layer perfrom wrong
operations.

gonzoleeman:
I understand that, in some circumstances, it would be nice to know that when iscsid stops, it is all done with the requests it has received, and that when it starts up again, it'll start cleanly (i.e. no old requests being processed).  But that seems like a fundamental change to me, i.e. it would change the default behavior. And some may be depending on that default behavior.

I can see the value in this, but it seems like there should be a configuration variable to control this, with the default being the current behavior.

@cleech  ??? @mikechristie  ???

mikechristie:
What is doing the kill -19 and why?

It seems if you just kill iscsid, it exits, but the systemd iscsd.socket service still stays up and queues requests, and you hit this issue right? Why does your tool that does the kill, not also stop the socket service? Would that avoid the issue?

I think for the case where iscsid crashes and the iscsid.socket service stays running, then systemd restarts iscsid, then we want to continue to execute requests when iscsid is auto restarted. The user in this case does not expect failures.

For the non systemd case, then we don't have this issue right? The requests fail because iscsiadm's poll detects the disconnected socket to iscsid.

mikechristie:
Hey,

There is maybe one related issue for the systemd and non-systemd case. If the request made it iscsid and it's being executed when iscsid is shutdown, those requests are left running in iscsid. They could complete ok, but iscsiadm hangs or gets an error depending on if you are using systemd or not.

Do you want to make iscsid shutdwn more graceful? It could complete running requests and then flush requests in the socket.

This would handle the case where systemd's iscsid.socket is running, iscsid is killed, and we have requests in the socket, and it would work for the non-systemd case.

For the case where systemd's iscsid.socket is running, iscsid is killed and then iscsiadm is run, we would consider those requests for the new instance of iscsid and should be executed like normal when iscsid is restarted. Note that, right now you probably want to kill these commands because if you updated iscsid at this time, the iscsiadm/iscsd request structs might be different versions.

wenchao-hao:
> What is doing the kill -19 and why?
> 
> It seems if you just kill iscsid, it exits, but the systemd iscsd.socket service still stays up and queues requests, and you hit this issue right? Why does your tool that does the kill, not also stop the socket service? Would that avoid the issue?
> 

The phenomenon which requests left unhandled in iscsid.socket does possible to happen but with low possibility, kill -19  would stop iscsid, it is just a method to recurrent requests left unhandled in iscsid.socket, usually no one do this.  

> I think for the case where iscsid crashes and the iscsid.socket service stays running, then systemd restarts iscsid, then we want to continue to execute requests when iscsid is auto restarted. The user in this case does not expect failures.

If iscsid crashes, the owner of these residual requests(usually iscsiadm) might already exit and error code is returned. If a newly started iscsid executes these requests, the actually result would mismatch with what iscsiadm returned.

> 
> For the non systemd case, then we don't have this issue right? The requests fail because iscsiadm's poll detects the disconnected socket to iscsid.

Yes, this phenomenon just happened with iscsid.socket enabled.

wenchao-hao:
> Hey,
> 
> There is maybe one related issue for the systemd and non-systemd case. If the request made it iscsid and it's being executed when iscsid is shutdown, those requests are left running in iscsid. They could complete ok, but iscsiadm hangs or gets an error depending on if you are using systemd or not.
> 
> Do you want to make iscsid shutdwn more graceful? It could complete running requests and then flush requests in the socket.

I would like to do this. While just make iscsid shutdown more graceful seems can not fix residual requests totally. If iscsid is killed abnormally, such as killed by signal 9, the iscsid's shutdown flow would not be executed.
> 
> This would handle the case where systemd's iscsid.socket is running, iscsid is killed, and we have requests in the socket, and it would work for the non-systemd case.
> 
> For the case where systemd's iscsid.socket is running, iscsid is killed and then iscsiadm is run, we would consider those requests for the new instance of iscsid and should be executed like normal when iscsid is restarted. Note that, right now you probably want to kill these commands because if you updated iscsid at this time, the iscsiadm/iscsd request structs might be different versions.
> 
> Edit:
> 
> I wanted to clarify this last part. I think we should add a version to the requests and check that. If they don't match then it should be failed.

Maybe this PR has did this check.  This PR introduce a unique id of each iscsid, this unique id is init as a random value when iscsid started up. 
When iscsiadm send requests to iscsid, it get current iscsid's unique id first and fill it in requests, iscsid would check if the ID in requests is matching with iscsid's own unique id, only requests whose ID matches with iscsid's own unique id would be handled.

gonzoleeman:
On Dec 11, 2021, at 12:47 AM, Wenchao Hao ***@***.***> wrote:
> 
> 
> Hey,
> 
> There is maybe one related issue for the systemd and non-systemd case. If the request made it iscsid and it's being executed when iscsid is shutdown, those requests are left running in iscsid. They could complete ok, but iscsiadm hangs or gets an error depending on if you are using systemd or not.
> 
> Do you want to make iscsid shutdwn more graceful? It could complete running requests and then flush requests in the socket.
> 
> I would like to do this. While just make iscsid shutdown more graceful seems can not fix residual requests totally. If iscsid is killed abnormally, such as killed by signal 9, the iscsid's shutdown flow would not be executed.
> 
> 
I don’t see how this can work correctly in all cases. The problem is that completing requests may mean waiting a bit, if either we are very congested (heavy I/O occurring) or a target is in error recovery, shutting down can mean waiting for a sometimes-long timeout. We have to ask ourselves which is more important, finishing the I/O or shutting down. Of course right now we just either shut down or hang, so there’s room for improvement IMHO.
> This would handle the case where systemd's iscsid.socket is running, iscsid is killed, and we have requests in the socket, and it would work for the non-systemd case.
> 
> For the case where systemd's iscsid.socket is running, iscsid is killed and then iscsiadm is run, we would consider those requests for the new instance of iscsid and should be executed like normal when iscsid is restarted. Note that, right now you probably want to kill these commands because if you updated iscsid at this time, the iscsiadm/iscsd request structs might be different versions.
> 
> Edit:
> 
> I wanted to clarify this last part. I think we should add a version to the requests and check that. If they don't match then it should be failed.
> 
> Maybe this PR has did this check. This PR introduce a unique id of each iscsid, this unique id is init as a random value when iscsid started up.
> When iscsiadm send requests to iscsid, it get current iscsid's unique id first and fill it in requests, iscsid would check if the ID in requests is matching with iscsid's own unique id, only requests whose ID matches with iscsid's own unique id would be handled.
> 
> 

I think part of the problem here is trying to handle every case, including when systemd is used and when it is not used, and when errors or timeouts are occurring. I have focused on using systemd in the last few years for open-iscsi, since that is the vast majority of usage, and I thought those that use open-iscsi without systemd would suggest fixes that addressed issue they saw.

So I must again say that I don’t like the idea of adding an “iscsid identity” token to requests between iscsiadm and iscsid, since much of the time we actually do want iscsid to complete pending requests. IMHO if we upgrade open-iscsi structures and brake the APIs between iscsid and iscsiadm, we should reboot, especially if the root disc is on iscsi.

— 
Lee Duncan

wenchao-hao:
> On Dec 11, 2021, at 12:47 AM, Wenchao Hao ***@***.***> wrote: Hey, There is maybe one related issue for the systemd and non-systemd case. If the request made it iscsid and it's being executed when iscsid is shutdown, those requests are left running in iscsid. They could complete ok, but iscsiadm hangs or gets an error depending on if you are using systemd or not. Do you want to make iscsid shutdwn more graceful? It could complete running requests and then flush requests in the socket. I would like to do this. While just make iscsid shutdown more graceful seems can not fix residual requests totally. If iscsid is killed abnormally, such as killed by signal 9, the iscsid's shutdown flow would not be executed.
> I don’t see how this can work correctly in all cases. The problem is that completing requests may mean waiting a bit, if either we are very congested (heavy I/O occurring) or a target is in error recovery, shutting down can mean waiting for a sometimes-long timeout. We have to ask ourselves which is more important, finishing the I/O or shutting down. Of course right now we just either shut down or hang, so there’s room for improvement IMHO.

The root cause is that iscsid is a single thread. Both kernel netlink messages and iscsiadm requests are handled in iscsid's loop, this mechanism can not guarantee all netlink messages or iscsiadm requests been handled when iscsid exit abnormally.
When iscsid.socket is enabled, systemd would listen on domain ISCSIADM_ABSTRACT_NAMESPACE, OS would not clear ISCSIADM_ABSTRACT_NAMESPACE even if iscsid exited.
So if iscsid keeps this mechanism of single thread unchanged, we need to take another method to solve newly-started iscsid executing residual requests left in iscsid.socket.

> This would handle the case where systemd's iscsid.socket is running, iscsid is killed, and we have requests in the socket, and it would work for the non-systemd case. For the case where systemd's iscsid.socket is running, iscsid is killed and then iscsiadm is run, we would consider those requests for the new instance of iscsid and should be executed like normal when iscsid is restarted. Note that, right now you probably want to kill these commands because if you updated iscsid at this time, the iscsiadm/iscsd request structs might be different versions. Edit: I wanted to clarify this last part. I think we should add a version to the requests and check that. If they don't match then it should be failed. Maybe this PR has did this check. This PR introduce a unique id of each iscsid, this unique id is init as a random value when iscsid started up. When iscsiadm send requests to iscsid, it get current iscsid's unique id first and fill it in requests, iscsid would check if the ID in requests is matching with iscsid's own unique id, only requests whose ID matches with iscsid's own unique id would be handled.

> I think part of the problem here is trying to handle every case, including when systemd is used and when it is not used, and when errors or timeouts are occurring. I have focused on using systemd in the last few years for open-iscsi, since that is the vast majority of usage, and I thought those that use open-iscsi without systemd would suggest fixes that addressed issue they saw. 

If iscsid.socket is disabled, this issue would not exist. But we should not tells users not to enable iscsid.socket, so this issue should be resolved by open-iscsi.

So I must again say that I don’t like the idea of adding an “iscsid identity” token to requests between iscsiadm and iscsid, since much of the time we actually do want iscsid to complete pending requests. IMHO if we upgrade open-iscsi structures and brake the APIs between iscsid and iscsiadm, we should reboot, especially if the root disc is on iscsi. — Lee Duncan

I think it is not necessary to reboot whole system, just restart iscsid is ok too.

mikechristie:
> On Dec 11, 2021, at 12:47 AM, Wenchao Hao ***@***.***> wrote: Hey, There is maybe one related issue for the systemd and non-systemd case. If the request made it iscsid and it's being executed when iscsid is shutdown, those requests are left running in iscsid. They could complete ok, but iscsiadm hangs or gets an error depending on if you are using systemd or not. Do you want to make iscsid shutdwn more graceful? It could complete running requests and then flush requests in the socket. I would like to do this. While just make iscsid shutdown more graceful seems can not fix residual requests totally. If iscsid is killed abnormally, such as killed by signal 9, the iscsid's shutdown flow would not be executed.
> I don’t see how this can work correctly in all cases. The problem is that completing requests may mean waiting a bit, if either we are very congested (heavy I/O occurring) or a target is in error recovery, shutting down can mean waiting for a sometimes-long timeout. We have to ask ourselves which is more important, finishing the I/O or shutting down. Of course right now we just either shut down or hang, so there’s room for improvement IMHO.

If something has started the iscsid shutdown operation, then that means you would not wait on requests. You fail the running ones and queued up ones.

If iscsid is stopped abnormally, then:
1. If not using systemd, the requests will fail already because the socket to iscsid will close.
2. If using systemd then existing users might be having iscsid automatically restart. In this case, I was saying when iscsid starts back up we have to continue to execute those commands like we do now.

mikechristie:
> IMHO if we upgrade open-iscsi structures and brake the APIs between iscsid and iscsiadm, we should reboot, especially if the root disc is on iscsi.

Yeah, I'm with wenchao-hao in that I don't see why you especially need to reboot when doing root on iscsi. Are there known deps that could be interacting with iscsid directly that can't be restarted when doing root on iscsi?

However, I disagree with wenchao-hao about using the unique ID to detect API changes. It would of course work, but I think the other effects it has are not always wanted.

mikechristie:
> So I must again say that I don’t like the idea of adding an “iscsid identity” token to requests between iscsiadm and iscsid, since much of the time we actually do want iscsid to complete pending requests.

What are all the cases for/against the patch?

Against:
1. I think for my iscsid crash and auto restart case we want to continue to run commands.
2. ? Lee, what are your other cases?

For:
1. wenchao-hao's case where iscsid can restart and end up executing the request successfully, even though iscsiadm has returned a failure?

@wenchao-hao for this case when is iscsiadm returning failure? For login/logout we always wait on iscsid or the socket to disconnect right? So for the case you are trying to handle where the request is in the socket, iscsid exits, then restarts and executes the req, the kernel and tools will be in sync right? I mean iscsiadm will have waited on iscsid restarting.

Are the commands you are concerned about the ones that are currently using iscsid_exec_req()?

mikechristie:
Sorry. I didn't mean to close this issue.

Reopening.

mikechristie:
> If iscsid crashes, the owner of these residual requests(usually iscsiadm) might already exit and error code is returned. If a newly started iscsid executes these requests, the actually result would mismatch with what iscsiadm returned.

Hey @wenchao-hao 

Specifically, I'm not sure I am understanding this part of the problem or can't replicate it.

In your example commands, iscsid re-starts up, reads in the request and executes and then iscsiadm completes. iscsiadm is always waiting for commands like login/logout.

For commands like login and logout, iscsiadm is *not* doing something like returning a failure early, then iscsid re-starts back up and reads in the request and executes ok.

However for commands that use iscsid_exec_req I can see how this happens, but except for discovery_offload_sendtargets I don't think any of those change the system state so nothing will notice as far as I can tell.

wenchao-hao:
> > If iscsid crashes, the owner of these residual requests(usually iscsiadm) might already exit and error code is returned. If a newly started iscsid executes these requests, the actually result would mismatch with what iscsiadm returned.
> 
> Hey @wenchao-hao
> 
> Specifically, I'm not sure I am understanding this part of the problem or can't replicate it.
> 
> In your example commands, iscsid re-starts up, reads in the request and executes and then iscsiadm completes. iscsiadm is always waiting for commands like login/logout.
> 
> For commands like login and logout, iscsiadm is _not_ doing something like returning a failure early, then iscsid re-starts back up and reads in the request and executes ok.
> 
Yes, it seems ok for login and logout requests.

> However for commands that use iscsid_exec_req I can see how this happens, but except for discovery_offload_sendtargets I don't think any of those change the system state so nothing will notice as far as I can tell.

I would pefrorm more tests to check what effects would result if iscsid execute requests which left unhandled.

wenchao-hao:
I would close this PR and perform more tests of these scenario

PR #291, opened 2021-12-04
iscsistart fix and timeout cleanup

mikechristie:
This fixes a bug where iscsistart would exit before we've waited for the login to complete or fail. It also cleans up some related code.

This is a followup to:

https://github.com/open-iscsi/open-iscsi/pull/275

### DIFF
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -241,26 +241,18 @@ static int login_session(struct node_rec *rec)
 	/*
 	 * Need to handle race where iscsid proc is starting up while we are
 	 * trying to connect. Retry with exponential backoff, start from 50 ms.
-	 *
-	 * NOTE: another race condition can occur if the system is just coming
-	 * up, where our login request will be sent, but the login response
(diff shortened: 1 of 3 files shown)


usr/iscsistart.c
```diff
@@ -241,26 +241,18 @@ static int login_session(struct node_rec *rec)
...
+		/*
+		 * Once our event loop is up then we want to wait for the login
+		 * response. Either it logs in, we hit the login retries count,
+		 * or this program crashes, so there no need for the response
+		 * timeout.
+		 */
+		rc = iscsid_exec_req(&req, &rsp, 0, -1);
 		if (rc == 0) {
 			return rc;
-		} else if (rc == ISCSI_ERR_SESSION_NOT_CONNECTED ||
```
gonzoleeman, line 262: Why is the session-not-connected error no longer checked for?

gonzoleeman, line 262: Why are we going to stop checking for session-not-connected?

PR #292, opened 2021-12-04
Speed up iscsid resync operation

mikechristie:
Perform the iscsid session resync operations in parallel and do not wait for the relogin return value.

note:

These patches were made over

https://github.com/open-iscsi/open-iscsi/pull/291

and include those patches.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1969,27 +1969,13 @@ static int queue_session_login_task_retry(struct login_task_retry_info *info,
 	return 0;
 }
 
-static int
-sync_conn(iscsi_session_t *session, uint32_t cid)
-{
-	iscsi_conn_t *conn;
-	int rc;
-
-	rc = __session_conn_create(session, cid);
-	if (rc)
-		return rc;
-	conn = &session->conn[cid];
-
-	/* TODO: must export via sysfs so we can pick this up */
-	conn->state = ISCSI_CONN_STATE_CLEANUP_WAIT;
-	return 0;
-}
-
 int
 iscsi_sync_session(node_rec_t *rec, queue_task_t *qtask, uint32_t sid)
 {
-	iscsi_session_t *session;
+	char state_buff[SCSI_MAX_STATE_VALUE];
 	struct iscsi_transport *t;
+	iscsi_session_t *session;
+	iscsi_conn_t *conn;
 	int err;
 
 	session = session_find_by_sid(sid);
@@ -2011,11 +1997,10 @@ iscsi_sync_session(node_rec_t *rec, queue_task_t *qtask, uint32_t sid)
 		goto destroy_session;
 	}
 
-	session->r_stage = R_STAGE_SESSION_REOPEN;
-
-	err = sync_conn(session, 0);
+	err = __session_conn_create(session, 0);
 	if (err)
 		goto destroy_session;
+	conn = &session->conn[0];
 
 	/*
 	 * The caller only cares we have rebuilt our state, and can process
@@ -2025,7 +2010,34 @@ iscsi_sync_session(node_rec_t *rec, queue_task_t *qtask, uint32_t sid)
 	qtask->rsp.command = MGMT_IPC_SESSION_SYNC;
 	mgmt_ipc_write_rsp(qtask, ISCSI_SUCCESS);
 
-	log_debug(3, "Started sync iSCSI session %d", session->id);
+	/*
+	 * Older kernels do not support needs_cleanup and/or the conn state
+	 * field, so we have to force the relogin.
+	 */
+	if (iscsi_sysfs_conn_needs_cleanup(sid)) {
+		conn_debug(4, conn, "Conn needs recovery cleanup.");
+		goto reopen;
+	}
+
+	memset(state_buff, 0, SCSI_MAX_STATE_VALUE);
+	if (iscsi_sysfs_get_conn_state(state_buff, sid)) {
+		goto reopen;
+	}
+	conn_debug(4, conn, "Conn in %s state.\n", state_buff);
+
+	if (strcmp("up", state_buff))
+		goto reopen;
+
+	conn->state = ISCSI_CONN_STATE_LOGGED_IN;
+	session->r_stage = R_STAGE_NO_CHANGE;
+	log_info("connection%d:%d is operational after resync.", session->id,
+		conn->id);
+	return 0;
+
+reopen:
+	conn_debug(3, conn, "Started sync iSCSI session");
+	conn->state = ISCSI_CONN_STATE_CLEANUP_WAIT;
+	session->r_stage = R_STAGE_SESSION_REOPEN;
 	session_conn_reopen(&session->conn[0], STOP_CONN_RECOVER);
 
 	return 0;
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -263,7 +263,7 @@ __session_conn_create(iscsi_session_t *session, int cid)
 
 	conn->state = ISCSI_CONN_STATE_FREE;
 	conn->session = session;
-	actor_init(&conn->login_timer, iscsi_login_timedout, NULL);
+	actor_init(&conn->login_timer, iscsi_login_timedout, conn);
 	/*
 	 * TODO: we must export the socket_fd/transport_eph from sysfs
 	 * so if iscsid is resyncing up we can pick that up and cleanup up
@@ -351,7 +351,6 @@ __session_create(node_rec_t *rec, struct iscsi_transport *t, int *rc)
 
 	INIT_LIST_HEAD(&session->list);
 	session->t = t;
-	session->reopen_qtask.mgmt_ipc_fd = -1;
 	session->id = INVALID_SESSION_ID;
 	session->use_ipc = 1;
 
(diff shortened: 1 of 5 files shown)


gonzoleeman:
Mike: I'm reviewing your changes.

mikechristie:
No problem. It's really low priority. If you are busy, feel free to wait until after winter break.

usr/iscsid.c
```diff
@@ -272,25 +275,45 @@ static int sync_session(__attribute__((unused))void *data,
 	/* multiple drivers could be connected to the same portal */
 	if (!iscsi_match_session(&rec, info))
 		return -1;
+
+	req = calloc(1, sizeof(*req));
+	if (!req)
+		return 0;
```
gonzoleeman, line 281: Should this be returning -1 or -ENOMEM?

usr/iscsid.c
```diff
@@ -590,10 +613,12 @@ int main(int argc, char *argv[])
...
 		pid = fork();
 		if (pid == 0) {
+			LIST_HEAD(sync_list);
 			int nr_found; /* not used */
 			/* child */
-			/* TODO - test with async support enabled */
-			iscsi_sysfs_for_each_session(NULL, &nr_found, sync_session, 0);
+			iscsi_sysfs_for_each_session(&sync_list, &nr_found,
+						     init_sync_session_req, 0);
+			sync_sessions(&sync_list);
```
gonzoleeman, line 621: I suppose you can optimize this by adding "if (nr_found)" before the call to sync_sessions(), correct? Since, usually, there will be none to sync. Not sure I feel strongly about this, though. :)

usr/iscsid.c
```diff
@@ -272,25 +275,45 @@ static int sync_session(__attribute__((unused))void *data,
...
+	return 0;
+}
+
+static void sync_sessions(struct list_head *sync_list)
+{
+	struct iscsid_async_req *req, *tmp;
+	iscsiadm_rsp_t rsp;
+	int retries = 0, rc;
+
+	list_for_each_entry_safe(req, tmp, sync_list, list) {
```
gonzoleeman, line 304: Should you not reset retries to zero for each entry in the list, rather than having 30 retries total for all items on the list, no matter how many?

gonzoleeman (changes requested):
Very nice work, as usual. Thank you! Just a couple of suggestions/questions.

usr/iscsid.c
mikechristie: Yeah, will fix.

usr/iscsid.c
mikechristie: I'll the check. We could have iscsi_sysfs_count_sessions return > 0, but then init_sync_session_req didn't set anything up.

usr/iscsid.c
mikechristie: We don't need to reset it. We actually might not even need it. I noticed the following while working on that code.

Originally, we were only handling a race where the fist request could be sent from this process, but main() was still doing the operations after the sync session fork, but had not yet done mgmt_ipc_listen and opened the iscsid socket and started listening on it.

However, I don't think that race will happen. Either I messed up and that race could never happen or someone fixed the code.

Right now, I think main() will do:

1. mgmt_ipc_listen(). That will then setup iscsid to start to listen on the iscsid socket.
2. We then fork the sync session process.
3. The sync session process might send a request.
4. If 3 happens before main does event_loop(), I think we will be ok because we already started listening on the socket in #1.

Note that for iscsistart, I think we can still race because we do the fork before mgmt_ipc_listen. I think we can change iscsistart so that it works like iscisd and then delete all that ISCSI_ERR_ISCSID_NOTCONN retry code in login_session.

I was thinking I'll wait on that GH issue about systemd and socket handling to shake out then modify this code accordingly if we some changes for that.

mikechristie:
> @mikechristie , where you going to update your submission?

Yeah, I'm waiting on our internal team to do some testing with it.

mikechristie:
@gonzoleeman I updated new patches with your comment handled, but don't merge them yet. I added another patch " iscsi initiator: avoid relogin during sync up ) to try and avoid relogin completely, but it requires a kernel patch that I want to discuss/review on the lists.

mikechristie:
These patches are hitting coredump:

https://github.com/open-iscsi/open-iscsi/issues/315

wenchao-hao:
> These patches are hitting coredump:
> 
> #315

@mikechristie  The coredump issue is closed now, the root reason is I did not make clean.

usr/initiator.c
```diff
@@ -2025,7 +2010,34 @@ iscsi_sync_session(node_rec_t *rec, queue_task_t *qtask, uint32_t sid)
...
+	 * Older kernels do not support needs_cleanup and/or the conn state
+	 * field, so we have to force the relogin.
+	 */
+	if (iscsi_sysfs_conn_needs_cleanup(sid)) {
+		conn_debug(4, conn, "Conn needs recovery cleanup.");
+		goto reopen;
+	}
+
+	memset(state_buff, 0, SCSI_MAX_STATE_VALUE);
+	if (iscsi_sysfs_get_conn_state(state_buff, sid)) {
```
gonzoleeman, line 2023: don't really need curly braces around a single goto. Just a nit ...

gonzoleeman (approved):
Just one tiny nit this time -- not enough of an issue for me to say no.

gonzoleeman:
I'm ready to merge this, unless there is still something pending I don't see ...

mikechristie:
I want to re-review this based on some issues I've found with the in-kernel cleanup patch and my patches to fix that and the bugs we've hit with offload.

gonzoleeman:
> I want to re-review this based on some issues I've found with the in-kernel cleanup patch and my patches to fix that and the bugs we've hit with offload.

No problem.

gonzoleeman:
@mikechristie -- did you get time to review this?

wenchao-hao:
Hi Mike and Lee, when would this PR merged?

I recently meet another issue which can be addressed by this PR. The issue is iscsid.service restart timeout because there are too many sessions to recovery. The root reason is iscsid sync session one by one when restarted, we would take a long time to recovery these sessions.

mikechristie:
I'll try to work on next week.

If not then, definitely will work on it between Dec 26 - Dec 30 because oracle is shutdown and I'll have nothing to do but look at iscsi patches :)

wenchao-hao:
> I'll try to work on next week.
> 
> If not then, definitely will work on it between Dec 26 - Dec 30 because oracle is shutdown and I'll have nothing to do but look at iscsi patches :)

It looks preety  good to me. I am doing some tests on it, I would focus on the iscsid restart to check if there are some conditions which was not covered..

mikechristie:
@wenchao-hao What part of this patchset do you need for your patches?

Patches:
0004-iscsi-sysfs-add-helpers-to-get-conn-state-and-cleanu.patch
0005-iscsi-initiator-avoid-relogin-during-sync-up.patch

- I'm doing to drop these from the patchset because it requires a kernel change. However, this is my preferred fix for speeding up iscsid resync, so I will make a new patchset for them when the kernel part is done.

Patches:
0003-iscsid-batch-sync-requests.patch

- I want to drop this patch. It turns out that it does make iscsid restart really fast, but some systems can't handle the flood of relogins so well. If we do something like patch 4 and 5 above, then we just don't have to relogin so that approach is fast and less destabilizing.

Patches:
0001-iscsid-simplify-qtask-handling.patch
0002-iscsid-speed-up-resync.patch
- I only needed these patches for patch3. Were these the patches you needed? If so, I'm ok with them. patch1 just needs to be rebased. I can break these 2 patches out into a new patchset if you needed them.

wenchao-hao:
The other patches also fixed some issues I meet.

Patches:
0003-iscsid-batch-sync-requests.patch
0002-iscsid-speed-up-resync.patch

Fix the issue of iscsid.service failed to restart if we have a massive number of session to sync.

It is easy to recurrent the issue:

1. discovery one iscsi node;
2. set the node's nr_sessions to a large number, for example 128
3. login this node, 128 sessions would be created
4. restart iscsid.service, the restart operations would failed with timeout

Your patches speed up the session sync flow, so the issue is fixed.
If you they are not going to be merged, we need another way to speed
iscsid's sync operations when restart.

mikechristie:
1. I was saying in one of the previous comments that for the normal case, we will speed up iscsid restart by just not relogging in the session. If the session is logged in, then we just re-build the state, so it's quick.

2. For the case where the sessions are down, and we do need to relogin, then you are right and we need something. I guess my patches for this case are no worse than if iscsid was already running and already trying to relogin N sessions in parallel.

Let me finish 1 and then we can do 1 and 2 together.

mikechristie:
> > 1. I was saying in one of the previous comments that for the normal case, we will speed up iscsid restart by just not relogging in the session. If the session is logged in, then we just re-build the state, so it's quick.
> 
> It seems just speed up the flow of recover session would not fix issue of iscsid.service failed to restart when we have a massive number of session to sync when iscsid restart, the issue would always exist if we wait all sessions to recover before notify systemd service is ready. But I still looking forword to your patches.

Let me make sure we are talking about the same thing.

You said there is an issue where you hit the systemd iscsid timeout because iscsid does not report it's ready within some timeout value, right? I agree this can happen because we currently wait for the session to relogin and on top of that there is a 2 second delay from the default time to wait, so 128 sessions is going to take 2+ minutes.

I was saying for the normal case we don't need to relogin the session. We only need to read in the sysfs and idbm info and re-create the session struct in userspace like in this patch

https://github.com/open-iscsi/open-iscsi/pull/292/commits/61bfeedd0644c3a576f108bff86c935bd7568f3a

This should be quick for even 1K sessions since it's just reading files, so we should not be hitting the iscsid systemd timeout for the normal case I thought. Are you saying that's not right?

PR #295, opened 2022-01-10
iscsi_err: Add iscsid request timed out error messages

wenchao-hao:
Commit 67eb8b2 added error code ISCSI_ERR_REQ_TIMEOUT.
While did not update array iscsi_error_list which referenced
in iscsi_err_to_str(). Which would access a invalid memory with
when variable of iscsi_err_to_str() is ISCSI_ERR_REQ_TIMEOUT.


mikechristie (commented):
Looks good to me. Thanks for finding and fixing this for me.

gonzoleeman:
Thank you!

PR #297, opened 2022-01-25
Fix issues discovered by gcc12

gonzoleeman:
Gcc-12 caught a few errors in the code where we were checking
for an array being empty incorrectly, so this commit fixes those.


gonzoleeman:
Thank you to Martin Liska for the heads up and for testing the code.

PR #298, opened 2022-01-28
Fix wrong install_systemd destination path

wenchao-hao:
systemd files should be installed to path "/usr/lib/systemd/system", 
while previous implement installs these files to "/usr/lib/systemd/system/system".

Following error would reported by systemd when start iscsid.service:
```
root@fedora ~:# systemctl status iscsid.service 
Unit iscsid.service could not be found.
```


gonzoleeman:
Thank you for catching and fixing this.

PR #300, opened 2022-02-01
actor: introduce thread name which would be printed to log.

wenchao-hao:
Previous actor log makes it hard to figure out what is going on in the thread, such as:

```deleting a scheduled/waiting thread!```

We only know a thread is deleted, while no other info like what this thread is about to do and which thread it is. Logs like following seems better:

```deleting a scheduled/waiting thread 01111e48: session_conn_error !```

It tells thread with id 01111e48, which is going to executed function "session_conn_error" is deleted.

This PR added a name to each thread and print thread name in log.

### CODE REVIEW

**Comment by wenchao-hao** [comment]
@gonzoleeman @mikechristie @cleech 
friendly ping...

**Comment by gonzoleeman** [comment]
I will review soon. 

--
Lee-Man Duncan

File: usr/actor.c [hunk 1]
```diff
@@ -168,8 +171,9 @@ actor_schedule_private(actor_t *thread, uint32_t delay_secs, int head)
 		// don't do anything
 		break;
 	case ACTOR_INVALID:
-		log_error("BUG: Trying to schedule a thread that has not been "
-			  "setup. Ignoring sched.");
+		log_error("BUG: Trying to schedule a thread %p:%s"
+			  "that has not been setup. Ignoring sched.\n",
+			  thread, thread->name);
```
cleech, line 180: In this case, it's checking for a schedule call on an actor_t that hasn't been initialized with actor_init (but apparently trusting that actor_t is zeroed so that state == ACTOR_INVALID == 0).  I don't know that it makes sense to trust thread->name in that case.  If the entire struct is zeroed it wouldn't matter, but it's never going to be trustworthy.

usr/actor.c
```diff
@@ -45,6 +45,10 @@ actor_time_left(actor_t *thread, time_t current_time)
 void
 __actor_init(actor_t *thread, void (*callback)(void *), void *data)
 {
+	if (thread->state != ACTOR_INVALID)
+		log_error("bug:thread %p:%s has already been inited",
```
cleech, line 49: Please have this say "initialized" instead of "inited"

usr/initiator_common.c
```diff
@@ -273,12 +273,16 @@ static int host_set_param(struct iscsi_transport *t,
...
+	char *buff = NULL;
 
-	if (type == ISCSI_STRING)
-		log_debug(3, "%s", value ? (char *)value : "NULL");
-	else
-		log_debug(3, "%u", *(uint32_t *)value);
+	if (type == ISCSI_STRING) {
+		buff = value ? (char *)value : "NULL";
+	} else {
+		buff = malloc(12);
```
cleech, line 281: is this malloced memory being leaked?

cleech: For such a small buffer, I think I'd rather see this on the stack to avoid needing to free it manually.

Or just put two complete log_debug calls in the two branches and do away with the snprintf and buff.
```
if (type == ISCSI_STRING)
    log_debug(3, "set operational parameter %d to %s", param, value ? (char *)value : "NULL");
else
    log_debug(3, "set operational parameter %d to %u", param, *(uint32_t *)value);
```

This looks like a victim of removing the manual newlines and making each log call append them.

cleech (changes requested):

usr/actor.c
wenchao-hao: > In this case, it's checking for a schedule call on an actor_t that hasn't been initialized with actor_init (but apparently trusting that actor_t is zeroed so that state == ACTOR_INVALID == 0). I don't know that it makes sense to trust thread->name in that case. If the entire struct is zeroed it wouldn't matter, but it's never going to be trustworthy.

Yes, here should not use thread->name. I would modify here to print thread address only.

usr/actor.c
wenchao-hao: > Please have this say "initialized" instead of "inited"

done

usr/initiator_common.c
wenchao-hao: > For such a small buffer, I think I'd rather see this on the stack to avoid needing to free it manually.
> 
> Or just put two complete log_debug calls in the two branches and do away with the snprintf and buff.
> 
> ```
> if (type == ISCSI_STRING)
>     log_debug(3, "set operational parameter %d to %s", param, value ? (char *)value : "NULL");
> else
>     log_debug(3, "set operational parameter %d to %u", param, *(uint32_t *)value);
> ```
> 
> This looks like a victim of removing the manual newlines and making each log call append them.

done

wenchao-hao:
@cleech Thanks for your review, I have modifed code according to your suggestions.

gonzoleeman:
This looks ok to me. Thank you, and thanks to Chris for his review as well.

gonzoleeman:
@cleech I believe you have to clear your "changes requested" flag? Please feel free to merge the changes, as well.

cleech (approved):

PR #303, opened 2022-02-10
Make session and connect log helpful to locate problem

wenchao-hao:
There are many logs like following which are not helpful to
locate problem when there are multiple sessions:

"unsupported opcode 0xxxx"

We can not know which connection or session the log belongs to,
so this PR find these logs and add session id, connection id
for them.

This would be helpful to locate problems according to logs.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -99,7 +99,7 @@ static void iscsi_ev_context_free(iscsi_conn_t *conn)
 
 		if (conn->context_pool[i]->allocated)
 			/* missing flush on shutdown */
-			log_error("BUG: context_pool leak %p",
+			conn_error(conn, "BUG: context_pool leak %p",
 				  conn->context_pool[i]);
 		free(conn->context_pool[i]);
 	}
@@ -127,7 +127,7 @@ iscsi_ev_context_get(iscsi_conn_t *conn, int ev_size)
 			/* some callers abuse this pointer */
 			ev_context->data = (void *)ev_context +
 					sizeof(struct iscsi_ev_context);
-			log_debug(7, "get ev context %p",
+			conn_debug(7, conn, "get ev context %p",
 				  &ev_context->actor);
 			return ev_context;
 		}
@@ -162,8 +162,7 @@ __login_response_status(iscsi_conn_t *conn,
 	case LOGIN_REDIRECTION_FAILED:
 		return CONN_LOGIN_RETRY;
 	default:
-		log_error("Login error (Login status %d) on conn %d", conn->id,
-			  login_status);
+		conn_error(conn, "Login error (Login status %d)", login_status);
 		break;
 	}
 
@@ -193,9 +192,8 @@ __check_iscsi_status_class(iscsi_session_t *session, int cid,
 			conn->failback_saddr = conn->saddr;
                         return CONN_LOGIN_IMM_REDIRECT_RETRY;
 		default:
-			log_error("conn %d login rejected: redirection "
-			        "type 0x%x not supported",
-				conn->id, status_detail);
+			conn_error(conn, "login rejected: redirection "
+			        "type 0x%x not supported", status_detail);
 			return CONN_LOGIN_RETRY;
 		}
 	case ISCSI_STATUS_CLS_INITIATOR_ERR:
@@ -206,13 +204,13 @@ __check_iscsi_status_class(iscsi_session_t *session, int cid,
 				session->id);
 			return CONN_LOGIN_AUTH_FAILED;
 		case ISCSI_LOGIN_STATUS_TGT_FORBIDDEN:
-			log_error("conn %d login rejected: initiator "
-			       "failed authorization with target", conn->id);
+			conn_error(conn, "login rejected: initiator "
+			       "failed authorization with target");
 			return CONN_LOGIN_AUTH_FAILED;
 		case ISCSI_LOGIN_STATUS_TGT_NOT_FOUND:
-			log_error("conn %d login rejected: initiator "
+			conn_error(conn, "login rejected: initiator "
 			       "error - target not found (%02x/%02x)",
-			       conn->id, status_class, status_detail);
+			       status_class, status_detail);
 			return CONN_LOGIN_FAILED;
 		case ISCSI_LOGIN_STATUS_NO_VERSION:
 			/*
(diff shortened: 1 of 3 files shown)


wenchao-hao:
@gonzoleeman @mikechristie @cleech Request  code review

gonzoleeman (approved):
Not sure why you didn't create a conn_warn(), except that each usage of log_warn seems a little different, so that's ok with me. Also, why not use "conn_error()" in mgmt_ipc.c? Perhaps because you don't include the correct dot-h file here? These are both nits IMHO.

Lastly, in the future, when you are messing with message strings that are currently broken over 2 lines, please consider joining them together. Split messages are a bit harder to find using grep, and gone are the days when we had 80-column terminals.

gonzoleeman:
I'll wait at least for Chris' review before proceeding. I did not list Mike as a requested reviewer since he's busy with other things these days, but if he gets time to review, that's great.

gonzoleeman:
@cleech ?? Want to review?

usr/initiator.c
```diff
@@ -453,7 +448,7 @@ static void iscsi_flush_context_pool(struct iscsi_session *session)
 static void
 __session_destroy(iscsi_session_t *session)
 {
-	log_debug(1, "destroying session");
+	log_debug(1, "destroying session %d", session->id);
```
cleech, line 451: "session %d" is inconsistent with everything added to this point as "session:%d"

usr/initiator.c
```diff
@@ -889,7 +883,7 @@ static void session_conn_error(void *data)
 	switch (error) {
 	case ISCSI_ERR_INVALID_HOST:
 		if (session_conn_shutdown(conn, NULL, ISCSI_SUCCESS))
-			log_error("BUG: Could not shutdown session.");
+			log_error("BUG: Could not shutdown session%d.", sid);
```
cleech, line 886: "session%d" is inconsistent with use of "session:%d" previously

usr/initiator.c
```diff
@@ -1159,13 +1155,13 @@ static void iscsi_stop(void *data)
 
 	rc = session_conn_shutdown(conn, conn->logout_qtask, ISCSI_SUCCESS);
 	if (rc)
-		log_error("BUG: Could not shutdown session.");
+		log_error("BUG: Could not shutdown session%d.", sid);
```
cleech, line 1158: another "session%d" vs "session:%d"

usr/initiator.c
```diff
@@ -1244,7 +1241,7 @@ static void iscsi_recv_async_msg(iscsi_conn_t *conn, struct iscsi_hdr *hdr)
 		break;
 	case ISCSI_ASYNC_MSG_PARAM_NEGOTIATION:
 		log_warning("Received async event param negotiation, "
-			    "dropping session");
+			    "dropping session %d", session->id);
```
cleech, line 1244: this one has a space

usr/initiator.c
```diff
@@ -1223,14 +1219,15 @@ static void iscsi_recv_async_msg(iscsi_conn_t *conn, struct iscsi_hdr *hdr)
...
+			   session->id, conn->id);
 		if (iscsi_send_logout(conn))
-			log_error("Could not send logout in response to"
+			conn_error(conn, "Could not send logout in response to"
 				 "logout request aen");
 		break;
 	case ISCSI_ASYNC_MSG_DROPPING_CONNECTION:
-		log_warning("Target dropping connection %u, reconnect min %u "
-			    "max %u", ntohs(async_hdr->param1),
+		log_warning("Target dropping connection%d:%d %u, reconnect min %u "
```
cleech, line 1229: Why not make a conn_warning if we're adding error and debug?

usr/mgmt_ipc.c
```diff
@@ -411,8 +411,14 @@ mgmt_ipc_write_rsp(queue_task_t *qtask, int err)
 	}
 
 	qtask->rsp.err = err;
-	if (write(qtask->mgmt_ipc_fd, &qtask->rsp, sizeof(qtask->rsp)) < 0)
-		log_error("IPC qtask write failed: %s", strerror(errno));
+	if (write(qtask->mgmt_ipc_fd, &qtask->rsp, sizeof(qtask->rsp)) < 0) {
+		if (qtask->conn && qtask->conn->session)
+			log_error("connect%d:%d IPC qtask write failed: %s",
```
cleech, line 416: Why not use conn_error here?  "connect" vs. "connection"

cleech:
In general I'm OK with this.  The last commit should probably use the conn_error macro and certainly needs to say "connection" instead of "connect".

I started flagging open coded session id use, but in general there's an inconsistent use of "session%d" vs "session:%d" vs "session %d" that I'd like to see resolved to a single format for at least the logging being touched here.

Should the conn format be "connection%d:%d" with no space or punctuation between the word connection and the session id?

A lot of this is superfluous just because of implementation details, but I'm OK with being extra careful in the logging in case something were to ever change (conn->session is never NULL, ISCSI_CONN_MAX is 1 so conn->id is always 0)

usr/initiator.c
wenchao-hao: done

usr/initiator.c
wenchao-hao: done

usr/mgmt_ipc.c
wenchao-hao: done

wenchao-hao:
> In general I'm OK with this. The last commit should probably use the conn_error macro and certainly needs to say "connection" instead of "connect".
> 
> I started flagging open coded session id use, but in general there's an inconsistent use of "session%d" vs "session:%d" vs "session %d" that I'd like to see resolved to a single format for at least the logging being touched here.
> 
> Should the conn format be "connection%d:%d" with no space or punctuation between the word connection and the session id?

The connection%d:%d style is to keep same with linux kernel logs.

> 
> A lot of this is superfluous just because of implementation details, but I'm OK with being extra careful in the logging in case something were to ever change (conn->session is never NULL, ISCSI_CONN_MAX is 1 so conn->id is always 0)

Yes, I noticed the conn->id should always be 0, but I did not know why, maybe a historical issue?

gonzoleeman:
> On Feb 25, 2022, at 8:37 AM, Wenchao Hao ***@***.***> wrote:
> 
> 
> In general I'm OK with this. The last commit should probably use the conn_error macro and certainly needs to say "connection" instead of "connect".
> 
> I started flagging open coded session id use, but in general there's an inconsistent use of "session%d" vs "session:%d" vs "session %d" that I'd like to see resolved to a single format for at least the logging being touched here.
> 
> Should the conn format be "connection%d:%d" with no space or punctuation between the word connection and the session id?
> 
> The connection%d:%d style is to keep same with linux kernel logs.
> 
> A lot of this is superfluous just because of implementation details, but I'm OK with being extra careful in the logging in case something were to ever change (conn->session is never NULL, ISCSI_CONN_MAX is 1 so conn->id is always 0)
> 
> Yes, I noticed the conn->id should always be 0, but I did not know why, maybe a historical issue?
> 
> 

The iscsi protocol allows for multiple connections per session (MCS), but it is not practical to implement this on Linux. I _believe_ the code was written thinking that we might someday support MCS, hence an array of connections.

— 
Lee D

cleech:
Thank you for the change, this looks OK to me

PR #309, opened 2022-02-24
Do not allow NOP-outs when running in initrd

gonzoleeman:
When open-iscsi is used for the root disc, it's
run at boot time in the initrd/initramfs. As the
documentation for open-iscsi discusses, NOPs are
a bad idea for this case. But it's difficult for
users to ensure NOPs are off when runnig in this
environment.

Fix this bug ensuring that NOPs are off (set to
zero) when running in initrd by checking for the
file /etc/initrd-release.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -249,6 +249,16 @@ __check_iscsi_status_class(iscsi_session_t *session, int cid,
 	return CONN_LOGIN_FAILED;
 }
 
+static bool
+in_initrd(void)
+{
+	bool res = false;
+	if (access("/etc/initrd-release", F_OK) == 0)
+		res = true;
+	log_debug(6, "in_initrd == %s", res ? "true" : "false");
+	return res;
+}
+
 static int
 __session_conn_create(iscsi_session_t *session, int cid)
 {
@@ -297,21 +307,28 @@ __session_conn_create(iscsi_session_t *session, int cid)
 	conn->auth_timeout = conn_rec->timeo.auth_timeout;
 
 	/* noop-out setting */
-	conn->noop_out_interval = conn_rec->timeo.noop_out_interval;
-	conn->noop_out_timeout = conn_rec->timeo.noop_out_timeout;
-	if (conn->noop_out_interval && !conn->noop_out_timeout) {
-		log_error("Invalid timeo.noop_out_timeout. Must be greater "
-			  "than zero. Using default %d.",
-			  DEF_NOOP_OUT_TIMEO);
-		conn->noop_out_timeout = DEF_NOOP_OUT_TIMEO;
-	}
+	if (in_initrd()) {
+		/* NOOPs disabled for firmware nodes */
+		conn->noop_out_interval = 0;
+		conn->noop_out_timeout = 0;
+	} else {
+		/* get the NOP values from the database for these nodes */
+		conn->noop_out_interval = conn_rec->timeo.noop_out_interval;
+		conn->noop_out_timeout = conn_rec->timeo.noop_out_timeout;
 
-	if (conn->noop_out_timeout && !conn->noop_out_interval) {
-		log_error("Invalid timeo.noop_out_interval. Must be greater "
-			  "than zero. Using default %d.",
-			  DEF_NOOP_OUT_INTERVAL);
-		conn->noop_out_interval = DEF_NOOP_OUT_INTERVAL;
+		/* validate the NOOP values */
+		if (conn->noop_out_interval && !conn->noop_out_timeout) {
+			log_error("Invalid timeo.noop_out_timeout. Must be greater than zero. Using default %d.",
+				  DEF_NOOP_OUT_TIMEO);
+			conn->noop_out_timeout = DEF_NOOP_OUT_TIMEO;
+		} else if (conn->noop_out_timeout && !conn->noop_out_interval) {
+			log_error("Invalid timeo.noop_out_interval. Must be greater than zero. Using default %d.",
+				  DEF_NOOP_OUT_INTERVAL);
+			conn->noop_out_interval = DEF_NOOP_OUT_INTERVAL;
+		}
 	}
+	log_debug(4, "Set connection NOOP-OUT interval/timeout to %u/%u",
+		  conn->noop_out_interval, conn->noop_out_timeout);
 
 	iscsi_copy_operational_params(conn, &session->nrec.session.iscsi,
 				      &conn_rec->iscsi);


gonzoleeman:
@cleech I would appreciate your review on this one, if you have time.

mikechristie:
What tool do you use for boot?
- For iscsistart based boots, it turns off nops by default already and you can override that to turn them on.

If you have setup dm-multipath and iscsi for the root, then you don't want iscsid to always override your nop settings during boot. You might have planned for the flakey networking during that time and wanted fast'ish boots, so turned on nops.

- For iscsiadm based boots, what do we do
1. For the case where you create a record on the fly in the initramfs then are you passing in the nop settings like the iscsistart/dracut case?
2. For the case where you have the user setup a node record then copy it to the initramfs then the nop setting is easy to set.
3. For the fw login case you need something.

gonzoleeman:
Mike:

Thanks for your reply!


On Feb 25, 2022, at 9:17 AM, mikechristie ***@***.***> wrote:
> 
> 
> What tool do you use for boot?
> 
> For iscsistart based boots, it turns off nops by default already and you can override that to turn them on.

We use iscsid/iscsiadm. If iscsid isn’t running then we can’t handle errors.
> If you have setup dm-multipath and iscsi for the root, then you don't want iscsid to always override your nop settings during boot. You might have planned for the flakey networking during that time and wanted fast'ish boots, so turned on nops.
> 
> For iscsiadm based boots, what do we do
> For the case where you create a record on the fly in the initramfs then are you passing in the nop settings like the iscsistart/dracut case?
> For the case where you have the user setup a node record then copy it to the initramfs then the nop setting is easy to set.
> For the fw login case you need something.

In our experience, with flakey network, NOPs don’t help. IMHO they only help when there are long periods of inactivity, and one wants to know if the network has gone down during that period, so as not to have to handle it later, when IO is needed. That doesn’t match with boot-time well.

I have never (yet) seen a case where iscsi NOPs have helped. Of course there could have been some cases, but I haven’t seen them. On the other hand, I see problems all the time with NOPs.

The initramfs does:

- iscsiadm -m fw — which creates NODE records for the target(s)

- iscsiadm -m fw -l -W — to login to all firmware nodes

The second step, the “firmware login”, fails to honor the noop_out_* values we have set in the node database.

But if we change the second steep to instead do “iscsiadm -m node -l -W”, then the NODE database is honored.

So perhaps I just need to figure out why the firmware login ignores our NODE entries. OR … I can modify our code to do the “node” login instead of the firmware login.

But I’m still not convinced that NOPs during boot time are good in any way.

— 
Lee

mikechristie:
> Mike: Thanks for your reply!
> On Feb 25, 2022, at 9:17 AM, mikechristie ***@***.***> wrote: What tool do you use for boot? For iscsistart based boots, it turns off nops by default already and you can override that to turn them on.
> We use iscsid/iscsiadm. If iscsid isn’t running then we can’t handle errors.
> If you have setup dm-multipath and iscsi for the root, then you don't want iscsid to always override your nop settings during boot. You might have planned for the flakey networking during that time and wanted fast'ish boots, so turned on nops. For iscsiadm based boots, what do we do For the case where you create a record on the fly in the initramfs then are you passing in the nop settings like the iscsistart/dracut case? For the case where you have the user setup a node record then copy it to the initramfs then the nop setting is easy to set. For the fw login case you need something.
> In our experience, with flakey network, NOPs don’t help. IMHO they only help when there are long periods of inactivity, and one wants to know if the network has gone down during that period, so as not to have to handle it 

Wait, I'm not sure if we are talking about the same thing. For the flakely network case and multipath setups, you want to just wait for the scsi cmd to timeout, then an abort to timeout before possibly switching paths?

The nop case helps with multipath when you can't have long delays for failover. Like if you have software on top of dm-multipath, then if there is a network issue you can't always (think of HA software using the device and it's timers) wait for the scsi cmd and abort timeout to fire before you switch paths (with just a couple paths this adds up). Note: Multipathd test IO might not kick in because queues are full on the initiator side and/or target side so a TUR is stuck and it's timer might not have even started.

For the boot case, if you are using dm-multipath, then you might have a similar goal where you want a speedy boot/failure in the face of network issues. I mean, I added the ability for specific customers and not just to do it :)

> later, when IO is needed. That doesn’t match with boot-time well. I have never (yet) seen a case where iscsi NOPs have helped. Of course there could have been some cases, but I haven’t seen them. On the other hand, I see problems all the time with NOPs. The initramfs does: - iscsiadm -m fw — which creates NODE records for the target(s) - iscsiadm -m fw -l -W — to login to all firmware nodes The second step, the “firmware login”, fails to honor 

It does this because it just reads sysfs like the "iscsiadm -m fw" does.

cleech:
I really don't like the implementation here, hard coding a check for a file that's specific to systemd-in-an-initramfs, even if the fallback case retains the current behavior.

So dracut has some calls that attempt to disable NOPs now that don't actually work?  The dracut module should get fixed.

We could probably do some other things to make it easier to control this stuff; add a command to copy out ibft records to files for editing before login, look at making some stuff changeable in a running session, get a --param option like iscsistart has on a login command (dracut switching to running iscsid in the initramfs lost some of the control that was there for just this sort of thing)

gonzoleeman:
On Feb 25, 2022, at 4:24 PM, Chris Leech ***@***.***> wrote:
> 
> 
> I really don't like the implementation here, hard coding a check for a file that's specific to systemd-in-an-initramfs, even if the fallback case retains the current behavior.
> 
> 

I kind of do like it, in that I can never see a reason for NOPs in the initrd/initramfs. NOPs IMHO can be useful when there are long periods of inactivity, which isn’t the case for the root disc during boot.
> So dracut has some calls that attempt to disable NOPs now that don't actually work? The dracut module should get fixed.
> 
> 

Already taken care of. Upstream just merged my pull request to remove the useless lines. Though I may have to submit another pull request to fully fix this issue.
> We could probably do some other things to make it easier to control this stuff; add a command to copy out ibft records to files for editing before login, look at making some stuff changeable in a running session, get a --param option like iscsistart has on a login command (dracut switching to running iscsid in the initramfs lost some of the control that was there for just this sort of thing)
> 
> 

I am next going to try a fix where the dracut code, which now does this:

* iscsid & (actually started by systemd)
* iscsiadm -m fw
* iscsiadm -m fw -l -W (send login requests but don’t wait for reply)
* iscsiadm -m node —op update —name … (two lines, to turn NOPs off)

Will become:

* iscsid &
* iscsiadm -m fw
* iscsiadm -m node —op update … (two lines, to turn off NOPs)
* iscsiadm -m node -l -W

In other words, turn off NOPs in the node database entry, then login via the NODE db, not firmware.

— 
Lee D

cleech:
It almost seems like it would be easiest for dracut to change these settings in /etc/iscsid.conf when the initramfs image is created.

gonzoleeman:
> It almost seems like it would be easiest for dracut to change these settings in /etc/iscsid.conf when the initramfs image is created.

:)

Of course I tried that. The problem is that when one uses "iscsiadm -m fw -l", open-iscsi does not use the node database at all, even if nodes are present for the listed targets.

I believe this sequence will work:
- discover the nodes (with "iscsiadm -m fw")
- update the NOP-out timeouts (set them to zero, i.e. "off")
- login using "iscsiadm -m node -T <tgt-name> -l" (or leave off the target name to login to all nodes, if there are two)

Using node mode to login *does* use the NODE database.

Other options would be to use iscsistart, or to modify iscsiadm to support passing in name/value pairs (as iscsistart does).

cleech:
> Of course I tried that. The problem is that when one uses "iscsiadm -m fw -l", open-iscsi does not use the node database at all, even if nodes are present for the listed targets.

Interesting, I realized it skipped the node records in favor of firmware data, but I thought it might have merged in defaults from iscsid.conf for all the settings we don't get out of an iBFT.  Thanks for being open to discussion on this.

gonzoleeman:
My testing so far has shown that the sequence of commands I suggested above for dracut to use actually do fix the problem, so I'm going to close this pull request.

Thanks you guys for the feedback and help.

PR #310, opened 2022-02-24
Fix iscsi-init.service start time

gonzoleeman:
The commit e6e53d6e60bf ("Fix iscsi-init so that it runs when
root writable") added "After=root.mount" to iscsi-init.service,
but that was an error, since that means waiting for "/root".

Remove that line and add the proper configuration so that
the iscsi-init service does not try to write to
/etc/iscsi/initiatorname.iscsi unless /etc/iscsi is
writable.

Fixes: e6e53d6e60bf

### DIFF
--- a/etc/systemd/iscsi-init.service
+++ b/etc/systemd/iscsi-init.service
@@ -2,7 +2,16 @@
 Description=One time configuration for iscsi.service
 ConditionPathExists=!/etc/iscsi/initiatorname.iscsi
 DefaultDependencies=no
-After=root.mount
+RequiresMountsFor=/etc/iscsi
+# systemd-remount-fs.service is optionally pulled in by
+# local-fs.target, don't start it here (no Wants=) but if
+# it's running wait for it to finish
+After=systemd-remount-fs.service
+
+[Install]
+# this ensures we are in the same transaction with
+# systemd-remount-fs.service
+WantedBy=systemd-remount-fs.service
 
 [Service]
 Type=oneshot


gonzoleeman:
@cleech , please review if you can, since we discussed this. Testing here shows it fixes the issue, though I'd still like to get ride of iscsi-init.service. At least this makes it work.

cleech:
I wasn't aware of RequiresMountsFor, seems good and works for me to wait for the root mount (-.mount).

I'm not sure about WantedBy=systemd-remount-fs.service, I think the Requires and After in iscsid.service should ensure that iscsi-init gets run if we need it.  If iscsid isn't enabled why bother generating an initiator name?

We could add a ConditionPathIsReadWrite=/etc/iscsi for extra checking as well.

gonzoleeman:
On Feb 25, 2022, at 3:58 PM, Chris Leech ***@***.***> wrote:
> 
> 
> I wasn't aware of RequiresMountsFor, seems good and works for me to wait for the root mount (-.mount).
> 
> I'm not sure about WantedBy=systemd-remount-fs.service, I think the Requires and After in iscsid.service should ensure that iscsi-init gets run if we need it. If iscsid isn't enabled why bother generating an initiator name?
> 
> We could add a ConditionPathIsReadWrite=/etc/iscsi for extra checking as well.
> 
I'm not sure about the added value of ConditionPathIsReadWrite. This doesn’t tell the unit to wait until it’s Read-Write. I believe it just fails the unit if the path is not writable.

I tried using just “-.mount”, but the problem is that systemd just waits for it to be mounted, not for it to be remounted read-write. So I had errors trying to write the initiator name file when I just used this.

— 
Lee

cleech:
My thought on ConditionPathIsReadWrite was just to get better logging of the failure, you're right in that in won't fix anything.

RequireMountsFor (or ConfigurationDirectory=iscsi) certainly seems like the right way to convert a needed path into whatever mount requirements.

Looking at the various early boot units included in systemd, it looks like After=systemd-remount-fs.service is the way to wait for writable mounts, I'm just questioning if that WantedBy is needed.

gonzoleeman:
> My thought on ConditionPathIsReadWrite was just to get better logging of the failure, you're right in that in won't fix anything.
> 

I guess I have no objecion to that, then. SUSE actually use the iscsi-gen-initiatorname command for this unit, and I have updated it to be more helpful and quit sooner when the filesystem is read-only (because of this issue). But having the Unit check wouldn't hurt, either, IMHO.

> RequireMountsFor (or ConfigurationDirectory=iscsi) certainly seems like the right way to convert a needed path into whatever mount requirements.

It was suggested by a user, and agreed to by our Systemd expert, so I can't take credit for it. I had never heard of it before.

> 
> Looking at the various early boot units included in systemd, it looks like After=systemd-remount-fs.service is the way to wait for writable mounts, I'm just questioning if that WantedBy is needed.

On this last point, here's what our systemd expert said:
> Possible drawback [to using systemd-remote-fs] is that iscsi-init.service may start even before
> systemd-remount-fs.service job is added to (systemd jobs) transaction.
> 
> It depends what pulls in iscsi-init.service, if it's local-fs.target,
> then it's fine since it'll be the same transaction with
> systemd-remount-fs.service.
> 
> Possible workaround would be to also add into iscsi-init.service:
> 
> 	[Install]
> 	\# just to ensure being in same transaction with
> 	\# systemd-remount-fs.service
> 	WantedBy=systemd-remount-fs.service
> 

But I have no evidence this is actually needed at this time, i.e. I've never had an error that having this WantedBy fixed.

gonzoleeman:
@cleech , if you have no objection, I'd like to merge this. If you object to the last bit (as discussed), let me know. We could possibly add that later, if needed.

cleech:
That's fine, this is certainly an improvement.

cleech (approved):

PR #313, opened 2022-02-28
REVIEW ONLY: dump of fedora modifications to etc/systemd

cleech:
Various modifications and hacks have accumulated in the Fedora/RHEL iscsi unit files that I should have synced with upstream ages ago.  Some of these were uglier before but maybe could be shared now, some of these differences are just out of date with recent upstream changes.

Some of this seems overly complicated, but it's dealt with a lot of picky requests about what should be happening at boot/service restart/shutdown.

I'm just dumping the current diff in here so I can use the Github review tools to go over this stuff.


gonzoleeman:
Very interesting, especially w/r/t iscsi boot. I should discuss with you some time.

PR #317, opened 2022-03-03
return ENOMEM if failed to get ev_context in iscsi_send_logout()

wenchao-hao:
If iscsi_send_logout() is called from iscsi_stop() or
session_login_task() and return 0, the next logout flow such as
session_conn_shutdown() depends on target's logout response and
logout timeout.

While if iscsi_send_logout() failed to trigger a logout timeout task,
and target's logout response absent, the logout flow would hang.

So we should return error if iscsi_send_logout() failed to trigger
a logout timeout task.

And add comment about handle logic if iscsi_send_logout() failed from
iscsi_recv_async_msg() by hand.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1150,8 +1150,11 @@ static void iscsi_stop(void *data)
 	iscsi_ev_context_put(ev_context);
 
 	if (!(conn->session->t->caps & CAP_LOGIN_OFFLOAD)) {
-		if (!iscsi_send_logout(conn))
+		rc = iscsi_send_logout(conn);
+		if (!rc)
 			return;
+		conn_error(conn, "Could not send logout pdu(%s) from iscsi_stop."
+			"Dropping session", strerror(rc));
 	}
 
 	rc = session_conn_shutdown(conn, conn->logout_qtask, ISCSI_SUCCESS);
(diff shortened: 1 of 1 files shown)


gonzoleeman:
Your changes look good, but it looks like the call to iscsi_send_logout() in session_logout_task() should perhaps also get the return value, and on failure add the "reason" to the error message?

wenchao-hao:
> Your changes look good, but it looks like the call to iscsi_send_logout() in session_logout_task() should perhaps also get the return value, and on failure add the "reason" to the error message?

OK, I would append a commit to print failure reason of iscsi_send_logout().

usr/initiator.c
```diff
@@ -2142,11 +2145,13 @@ int session_logout_task(int sid, queue_task_t *qtask)
...
 			/* unbind is not supported so just do old logout */
-			if (!iscsi_send_logout(conn))
+			rc = iscsi_send_logout(conn);
+			if (rc)
 				return ISCSI_SUCCESS;
 		}
 
-		conn_error(conn, "Could not send logout pdu. Dropping session");
+		conn_error(conn, "Could not send logout pdu(%s) from session_logout_task."
+			"Dropping session", strerror(rc));
```
gonzoleeman, line 2154: Nit: you have no space between your two text strings. In general, please do NOT break messages like this over two lines, as that makes it hard to search for (and read) messages.

usr/initiator.c
```diff
@@ -1150,8 +1150,11 @@ static void iscsi_stop(void *data)
 	iscsi_ev_context_put(ev_context);
 
 	if (!(conn->session->t->caps & CAP_LOGIN_OFFLOAD)) {
-		if (!iscsi_send_logout(conn))
+		rc = iscsi_send_logout(conn);
+		if (!rc)
 			return;
+		conn_error(conn, "Could not send logout pdu(%s) from iscsi_stop."
+			"Dropping session", strerror(rc));
```
gonzoleeman, line 1157: Nit: see below: no need to break this into two lines, and you missed a space between sentences because of it.

gonzoleeman (approved):

gonzoleeman:
@cleech if you have no objections we can merge this pull request, AFAIC

gonzoleeman:
Merged. Thank you.

PR #318, opened 2022-03-10
FW login allow nv pairs and set defaults

gonzoleeman:
Please review this pull request. It addresses an issue where firmware logins don't use the Node DB, so you are not able to set things like noop_out_timeout/interval, even if there are Node DB entries for these targets.

This change does 4 things:
1. Adds the ability to pass in name/value pairs for "iscsiadm -m fw -l  [-W]" calls, using the standard "-n"/"--name" and "-v"/"--value" command line options.
2. It sets the defaults for firmware login to NOP-OUTs being off, unless overridden with these new command-line options.
3. It adds info about name/value pairs for firmware logins to the iscsiadm.8 man page.
4. It does a somewhat large cleanup of the iscsiadm.8 man page, including removing duplicate spaces, cleaning up some working/grammar, and trying to be consistent about using italics (emphasis) for the names of command line arguments and options.

I would really love some feedback on this. As far as the firmware login fix, I can't imagine anyone will disagree with it.

I perhaps should have broken the man page fix into two commits: one that cleans things up, and then one that adds the name/value firmware info, but I found myself compelled to fix the man page issue and didn't want to backtrack. But if there is objection, I can backtrack as needed.

@cleech and @mikechristie I'd really love your input, as well as anyone else that would like to comment.

### DIFF
--- a/usr/iscsiadm.c
+++ b/usr/iscsiadm.c
@@ -114,7 +114,7 @@ static const struct verify_mode_t mode_paras[] = {
 	[MODE_SESSION] = {"session", "PiRdrmusonuSv", 1},
 	[MODE_HOST] = {"host", "CHdmPotnvxA", 0},
 	[MODE_IFACE] = {"iface", "HIdnvmPoCabci", 0},
-	[MODE_FW] = {"fw", "dmlW", 0},
+	[MODE_FW] = {"fw", "dmlWnv", 0},
 };
 
 static struct option const long_options[] =
@@ -165,10 +165,10 @@ iscsiadm -m discoverydb [-hV] [-d debug_level] [-P printlevel] [-t type -p ip:po
 [-o operation] [-n name] [-v value] [-lD]] \n\
 iscsiadm -m discovery [-hV] [-d debug_level] [-P printlevel] [-t type -p ip:port -I ifaceN ... [-l]] | [[-p ip:port] [-l | -D]] [-W]\n\
 iscsiadm -m node [-hV] [-d debug_level] [-P printlevel] [-L all,manual,automatic,onboot] [-W] [-U all,manual,automatic,onboot] [-S] [[-T targetname -p ip:port -I ifaceN] [-l | -u | -R | -s]] \
-[[-o  operation ] [-n name] [-v value]]\n\
-iscsiadm -m session [-hV] [-d debug_level] [-P  printlevel] [-r sessionid | sysfsdir [-R | -u | -s] [-o operation] [-n name] [-v value]]\n\
-iscsiadm -m iface [-hV] [-d debug_level] [-P printlevel] [-I ifacename | -H hostno|MAC] [[-o  operation ] [-n name] [-v value]] [-C ping [-a ip] [-b packetsize] [-c count] [-i interval]]\n\
-iscsiadm -m fw [-d debug_level] [-l] [-W]\n\
+[[-o operation ] [-n name] [-v value]]\n\
+iscsiadm -m session [-hV] [-d debug_level] [-P printlevel] [-r sessionid | sysfsdir [-R | -u | -s] [-o operation] [-n name] [-v value]]\n\
+iscsiadm -m iface [-hV] [-d debug_level] [-P printlevel] [-I ifacename | -H hostno|MAC] [[-o operation ] [-n name] [-v value]] [-C ping [-a ip] [-b packetsize] [-c count] [-i interval]]\n\
+iscsiadm -m fw [-d debug_level] [-l] [-W] [[-n name] [-v value]]\n\
 iscsiadm -m host [-P printlevel] [-H hostno|MAC] [[-C chap [-x chap_tbl_idx]] | [-C flashnode [-A portal_type] [-x flashnode_idx]] | [-C stats]] [[-o operation] [-n name] [-v value]] \n\
 iscsiadm -k priority\n");
 	}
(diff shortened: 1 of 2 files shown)


gonzoleeman:
@cleech @mikechristie Ping?

usr/iscsiadm.c
```diff
@@ -3013,8 +3013,78 @@ static int exec_fw_disc_op(discovery_rec_t *drec, struct list_head *ifaces,
...
+		 * ibft/iscsi_boot, so clear out values that might conflict
+		 * with user overrides
+		 */
+		if (!strcmp(param->name, IFACE_NETNAME)) {
+			/* overriding netname so MAC will be for old netdev */
+			memset(rec->iface.hwaddress, 0,
+				sizeof(rec->iface.hwaddress));
+		} else if (!strcmp(param->name, IFACE_HWADDR)) {
+			/* overriding MAC so netdev will be for old MAC */
+			memset(rec->iface.netdev, 0, sizeof(rec->iface.netdev));
```
mikechristie, line 3046: This should be hwaddress.

usr/iscsiadm.c
```diff
@@ -3013,8 +3013,78 @@ static int exec_fw_disc_op(discovery_rec_t *drec, struct list_head *ifaces,
...
+	list_for_each_entry(param, params, list) {
+		/*
+		 * user may not have passed in all params that were set by
+		 * ibft/iscsi_boot, so clear out values that might conflict
+		 * with user overrides
+		 */
+		if (!strcmp(param->name, IFACE_NETNAME)) {
+			/* overriding netname so MAC will be for old netdev */
+			memset(rec->iface.hwaddress, 0,
+				sizeof(rec->iface.hwaddress));
```
mikechristie, line 3043: This should be iface.netdev in the memset, or the comment is wrong, or I don't get it :)

mikechristie: I think it's the latter. I'm not sure what is going on for all the iface fields.

usr/iscsiadm.c
```diff
@@ -3013,8 +3013,78 @@ static int exec_fw_disc_op(discovery_rec_t *drec, struct list_head *ifaces,
...
+	/*
+	 * For root boot we could not change this in older versions so
+	 * if user did not override then use the defaults.
+	 *
+	 * Increase to account for boot using static setup.
+	 */
+	if (rec->session.initial_login_retry_max == -1)
+		rec->session.initial_login_retry_max = 30;
+
+	/* we used to not be able to answer so turn off */
```
mikechristie, line 3074: This was a iscsistart specific comment. When it was created we did nops in userspace so if we sent a nop off we probably would exit before we got a response. You would hit a bug like we fixed the other day. I think we also defaulted to nops off so that code in iscsistart is to keep that behavior (at some point the default was to turn them on and so we had to override in iscsistart).

For as long as we could do ibft/fw based boots we could do kernel based nops so you probably want to change the comment or drop it. Maybe just change the comment or add a mention in the git commit that we are turning off by default now to optimize for the single path case. If multipath users were relying on the defaults for fast boots during failures then they need to change them.

usr/iscsiadm.c
mikechristie: I remember this now :) I'm not sure if it's right for some cases now.

Can you make a helper function so it's not duplicated with iscsistart.

gonzoleeman:
On Mar 14, 2022, at 3:40 PM, mikechristie ***@***.***> wrote:
> 
> ﻿
> @mikechristie commented on this pull request.
> 
> In usr/iscsiadm.c:
> 
> > +	/* Must init this so we can check if user overrode them */
> +	rec->session.initial_login_retry_max = -1;
> +	rec->conn[0].timeo.noop_out_interval = -1;
> +	rec->conn[0].timeo.noop_out_timeout = -1;
> +	rec->session.scan = -1;
> +
> +	list_for_each_entry(param, params, list) {
> +		/*
> +		 * user may not have passed in all params that were set by
> +		 * ibft/iscsi_boot, so clear out values that might conflict
> +		 * with user overrides
> +		 */
> +		if (!strcmp(param->name, IFACE_NETNAME)) {
> +			/* overriding netname so MAC will be for old netdev */
> +			memset(rec->iface.hwaddress, 0,
> +				sizeof(rec->iface.hwaddress));
> I remember this now :) I'm not sure if it's right for some cases now.
> 
> Can you make a helper function so it's not duplicated with iscsistart.
> 

Yes. I expected such a request.

The two uses are slightly different so it’s nontrivial. :)
— 
Lee

usr/iscsiadm.c
mikechristie: > I remember this now :) I'm not sure if it's right for some cases now.
> 
> Can you make a helper function so it's not duplicated with iscsistart.

I want to change this comment after sleeping on it. It crept up in my dreams :(

I don't think we want this chunk and instead of the user passes them in fail. The iscsistart code is from a different time where we supported weird mix and match cases because offload boot and ibft/ppc boot were just coming out.

For the iscsiadm fw boot, we should start cleanly and not bring cruft over unless we need to. Since we currently only support booting from the interfaces that are presented in the iscsi_boot_sysfs interface (ibft/ppc fw/offload) then we should support it and not support/allow someone to override it.

So don't allow the user to specific those iface values.

gonzoleeman:
I will withdraw this pull request and submit another with an updated branch. I'll leave this open, for now, as a reference.

gonzoleeman:
Replaced by pull request #321

PR #321, opened 2022-03-16
Firmware login: add nv pairs and default values v2

gonzoleeman:
Updated from previous pull request #318 which will be replaced by this one.

Changes from that request:

- unrelated man page cleanup done separately, mostly (one small cleanup snuck in)
- removed ability to override interface values, instead now failing with an error message
- changed iscsistart-specific comment in new code to be firmware-login-specific
- Also removed validation of interface name changes, since they were node-specific and no longer needed, since we disallow interface value changes

Has been tested in Virtual Machine using iscsi multipath root disc.

### DIFF
--- a/doc/iscsiadm.8
+++ b/doc/iscsiadm.8
@@ -123,6 +123,10 @@ iscsiadm \- open-iscsi administration utility
 .IR debug_level ]
 .RB [ \-l ]
 .RB [ \-W ]
+.RB [ \-n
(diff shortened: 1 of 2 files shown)


gonzoleeman:
@mikechristie I believe I addresses your issues, if you get a chance to check.

mikechristie:
Looks ok to me.

PR #324, opened 2022-03-20
Clean up Makefile build system.

gonzoleeman:
A major overdue overhaul of the Make layout, including
moving lots of lower-level things out of the top-level
Makefile into the Makefiles for the appropriate subdirectory,
renaming the template files (where @SBINDIR@ gets translated)
so that their purpose is more obvious, making the build more
idempotent (no need to install a file if it's not newer).

Should be no functional change, as no script or C file changed.

Main difference from before: now the udev rules gets made and
installed by default (as part of "make"/"make install"), and
the iscsiuio daemon is now installed for "make install".

One should now be able to go into any subdirectory and type
"make"/"make install" to build and install the targets from
that directory, at least for the etc, utils, and usr
subdirectories.

This change is in preparation for more template translations
coming.


gonzoleeman:
Note: **SHOULD** be no functional change. Only change is in what happens for "make" and "make install", and only 2 changes there as far as what gets installed: now udev rules file and iscsiuio are installed as part of "make install". I can change that back if anybody has an objection ...

cleech (approved):
This all looks good to me.

wenchao-hao:
> One should now be able to go into any subdirectory and type
> "make"/"make install" to build and install the targets from
> that directory, at least for the etc, utils, and usr
> subdirectories.

@gonzoleeman 
Hey Lee, I found a small issue which it would fail to run make in usr subdirectories before running make in libopeniscsiusr, utils/sysdeps and utils/fwparam_ibft.

Maybe we should add recipe in usr/Makefile to make those  subdirectories strong dependence like following:

```
--- a/usr/Makefile
+++ b/usr/Makefile
@@ -56,7 +56,7 @@ PROGRAMS      = iscsid iscsiadm iscsistart
 PROGRAMS_DEST  = $(addprefix $(SBINDIR)/,$(PROGRAMS))
 
 # libc compat files
-SYSDEPS_SRCS = $(sort $(wildcard ../utils/sysdeps/*.o))
+SYSDEPS_SRCS = ../utils/sysdeps/sysdeps.o
 # sources shared between iscsid, iscsiadm and iscsistart
 ISCSI_LIB_SRCS = iscsi_util.o io.o auth.o iscsi_timer.o login.o log.o \
        iface.o idbm.o sysfs.o host.o session_info.o iscsi_sysfs.o \
@@ -93,6 +93,9 @@ $(SBINDIR):
 $(PROGRAMS_DEST): $(SBINDIR)/%: %
        $(INSTALL) -m 755 $? $@
 
+SYSDEPS_SRCS:
+       make -C ../utils/sysdeps
+
 clean:
        rm -f *.o $(PROGRAMS) .depend $(LIBSYS)
```

If we do not fix this issue, we need update the Makefile in top directory as following:

```
diff --git a/Makefile b/Makefile
index 2e2627d..89ad1bf 100644
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,9 @@ install_udev_rules:
        $(MAKE) $(MFLAGS) -C utils $@
 
 install_programs:
+       $(MAKE) $(MFLAGS) -C libopeniscsiusr SBINDIR=$(sbindir)
+       $(MAKE) $(MFLAGS) -C utils/sysdeps
+       $(MAKE) $(MFLAGS) -C utils/fwparam_ibft
        $(MAKE) $(MFLAGS) -C utils install
        $(MAKE) $(MFLAGS) -C usr install
```

gonzoleeman:
Hi @wenchao-hao  -- good point. One I struggle with. The goal of the make system is to have knowledge of how to build things in one place only. I do not want the formula for making systems to be in 3 or 4 places.

My philosophy has always been: if you want to build things in order, (1) use the top-level Makefile, or (2) know what you're doing. I do not want libopeniscsiusr, usr, and iscsiuio all knowing how build sysdeps. (I don't control iscsiuio though)

I will think about how to address this a bit. Please open a separate issue on this? Thank you.

PR #325, opened 2022-03-25
Make etc iscsi configurable

gonzoleeman:
This large commit adds "DBROOT=/some/dir" and "HOMEDIR=/some/other/dir" options to the Makefiles, so that files can be moved out of /etc/iscsi if desired. DBROOT is for the database files, which some distributions have already moved to /var/lib/iscsi -- so this might make that easier. HOMEDIR is where iscsid.conf and initiatorname.iscsi live. Both still default to /etc/iscsi, for now, though my hope is that the database files will moved to /var/lib/iscsi, by default, soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,16 +85,16 @@ clean:
 # note that make may still execute the blocks in parallel
 .NOTPARALLEL: install_user install_programs install_initd \
 	install_initd_redhat install_initd_debian \
-	install_etc install_iface install_doc install_iname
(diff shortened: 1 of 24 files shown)


gonzoleeman:
Note: the first two small commits just fix issues from the previous Makefile update that are unrelated to adding HOMEDIR and DBROOT, but needed to be done.

gonzoleeman:
Last call on reviews of this pull request ...

Issue #327, opened 2022-03-29
Failed to run make in usr subdirectories


wenchao-hao:
I found a small issue, we would fail to run make in usr subdirectories before running make in libopeniscsiusr, utils/sysdeps and utils/fwparam_ibft. The root reason is that programs in usr subdirectories depends on object files in other directories.

gonzoleeman:
I think this pull request of mine addresses your issue in a reasonable way. Now, in a clean tree, "make" in "usr" will succeed.

I believe we can close this issue once that pull request is merged.

gonzoleeman:
Marking as closed. Thank you for your issue and the review.

PR #328, opened 2022-03-31
More makefile cleanup

gonzoleeman:
This continues the Makefile/build cleanup of the last couple of commits. In particular, this fixes the issue when running "make" in the "usr" subdirectory. Before this, such a "make" could fail if the sysdeps or fwparam_ibft files hadn't been compiled first. Also, the fwparam_ibft stuff is cleaned up a little, as is usr/Makefile and Makefile.

Addresses github issue #327


wenchao-hao:
It looks good to me

PR #331, opened 2022-04-02
using ISCSI_UEVENT_DESTROY_SESSION_ASYNC to destoy session

### LINKED ISSUE #312: Use async session destroy nl cmd
Hey Lee,

If you are looking for work for your lots of targets work/talk, you should convert open-iscsi to use ISCSI_UEVENT_DESTROY_SESSION_ASYNC. 

The issue on the list is actually a massive number of devices instead of sessions/targets. There are 2k devices (40 sessions (20 per port), and 50 devices). The problem is that the queue operations like blocking and deletion are not free anymore. This conflicts with iscsid's design where it assumed most nl operations are non blocking (ignoring mem allocs) or are really short waits.

For session removal if the session is not logged in then session_logout_task skips the ubind call (async target removal) and proceeds to just call into the kernel to cleanup and free the iscsi related structs. If there are lots of devices, then scsi_transport_iscsi.c:iscsi_remove_session will have to loop over them and possibly wait for existing works that are also looping over them. This then blocks up the main thread in iscsid.

So we could use ISCSI_UEVENT_DESTROY_SESSION_ASYNC which it looks like it was added for this type of thing. When I was fixing the in-kernel recovery code I saw the patch and talked to the google devs. They said it was actually added because of an issue in sd's shutdown that was not scaling that Bart later fixed but it's similar as in an issue during removal that was because the non blocking assumption was broken at some point.

We could also just change session_logout_task to always call session_unbind. This has the nice side effect that if a cache sync is needed and we can relogin to the target, then it can get sent if we are able to relogin. It has the drawback, that if the user turned off the recovery timeout (set it to -1) then it will take the iscsiadm logout command longer to return than before, because instead of immediately cleaning up it will now have to wait for relogin timer to fire.

Alternatively, we can just give up on the non blocking assumption and stop doing the single process iscsid. This is nice, because the sd shutdown issue was probably a regression. Shutdown like queue operations were faster at one point. In the future other operations may also block.

Take your pick.

wenchao-hao:
The netlink assumed operations with kernel are fast, so the implement of __kipc_call() is a sync operation. It would wait for kernel's response once it sent the ISCSI_UEVENT_XXX.

While the ISCSI_UEVENT_DESTROY_SESSION might takes a long time to get a response from kernel, which would block iscsid.

Following is a scenario:
    
When connection is not in ISCSI_CONN_STATE_LOGGED_IN, session_unbind() would not be called. So when ISCSI_UEVENT_DESTROY_SESSION is sent to kernel, the kernel might perform unbind session, which is slow in     session with a massive number of devices.

So here we using ISCSI_UEVENT_DESTROY_SESSION_ASYNC other than ISCSI_UEVENT_DESTROY_SESSION to address it.

Fix #312


wenchao-hao:
@gonzoleeman @mikechristie I look forward for your review.

.gitignore
```diff
@@ -1,4 +1,6 @@
 *.o
+*.gcno
+*.gcda
 cscope.files
 cscope.out
 cscope.in.out
```
mikechristie, line 6: Drop this.

usr/netlink.c
```diff
@@ -443,7 +443,7 @@ kdestroy_session(uint64_t transport_handle, uint32_t sid)
 
 	memset(&ev, 0, sizeof(struct iscsi_uevent));
 
-	ev.type = ISCSI_UEVENT_DESTROY_SESSION;
+	ev.type = ISCSI_UEVENT_DESTROY_SESSION_ASYNC;
 	ev.transport_handle = transport_handle;
 	ev.u.d_session.sid = sid;
```
mikechristie, line 449: It's going to be more complicated then just switching the type.

1. We need to support older kernels that didn't have this cmd. So you would try async first. If that returns a not supported error then drop down to sync.
2. For async iscsid can't free eveything once the cmd returns from the kernel, or we will have a case where the kernel still thinks there is a session, but userspace does not. A user might then expect his LUNs and session are not in use and do other management ops like umount/reuse a LUN or relogin. They will then get errors but won't know why since iscsiadm's logout command returned success. We have to wait for the kernel to report the session is destroyed with this event in drivers/scsi/scsi_transport_iscsi.c:

void iscsi_free_session(struct iscsi_cls_session *session)
{
        ISCSI_DBG_TRANS_SESSION(session, "Freeing session\n");
        iscsi_session_event(session, ISCSI_KEVENT_DESTROY_SESSION);
        put_device(&session->dev);
}

In open-iscsi we just use this for this one weird driver, qla4xxx, in open-iscsi/user/initiator.c:iscsi_async_session_destruction and all we do is log a debug message. That driver works in a way where all the state is stored in the card's flash. So instead of have the node DB. It has a DB  of nodes in its flash that's read/written to, and open-iscsi just gets/sets that stuff.

For the non-qla4xxx case we have to modify the code so iscsi_async_session_destruction frees the userspace/iscsid session at that time.

So it will work more like the target/session unbind operation where session_unbind sends a NL cmd to tell the kernel to delete the iscsi target and it's LUNs from the session. The kernel returns immediately, and then does a:

        iscsi_session_event(session, ISCSI_KEVENT_UNBIND_SESSION);

to tell userspace when the target is deleted. open-iscsi then gets that event and proceeds to stop and teardown the conn/session.

wenchao-hao: Hi @mikechristie, thanks for your review

I want to do this as following:

1. add a new callback destroy_session_async in iscsi_ipc, which implements a ISCSI_UEVENT_DESTROY_SESSION_ASYNC, like following:

```
diff --git a/include/iscsi_if.h b/include/iscsi_if.h
index 22b7c09..50f3b1b 100644
--- a/include/iscsi_if.h
+++ b/include/iscsi_if.h
@@ -82,6 +82,7 @@ enum iscsi_uevent_e {
        ISCSI_UEVENT_LOGOUT_FLASHNODE_SID       = UEVENT_BASE + 30,
        ISCSI_UEVENT_SET_CHAP           = UEVENT_BASE + 31,
        ISCSI_UEVENT_GET_HOST_STATS     = UEVENT_BASE + 32,
+       ISCSI_UEVENT_DESTROY_SESSION_ASYNC = UEVENT_BASE +33,
        ISCSI_UEVENT_MAX                = ISCSI_UEVENT_GET_HOST_STATS,
 
        /* up events */
diff --git a/usr/iscsi_ipc.h b/usr/iscsi_ipc.h
index 47857dd..99e236f 100644
--- a/usr/iscsi_ipc.h
+++ b/usr/iscsi_ipc.h
@@ -78,6 +78,7 @@ struct iscsi_ipc {
                               uint32_t *hostno);
 
        int (*destroy_session) (uint64_t transport_handle, uint32_t sid);
+       int (*destroy_session_async) (uint64_t transport_handle, uint32_t sid);
 
        int (*unbind_session) (uint64_t transport_handle, uint32_t sid);
 
diff --git a/usr/netlink.c b/usr/netlink.c
index 22cad83..04edccc 100644
--- a/usr/netlink.c
+++ b/usr/netlink.c
@@ -456,6 +456,30 @@ kdestroy_session(uint64_t transport_handle, uint32_t sid)
        return 0;
 }
 
+static int
+kdestroy_session_async(uint64_t transport_handle, uint32_t sid)
+{
+       int rc;
+       struct iscsi_uevent ev;
+       struct iovec iov[2];
+
+       log_debug(7, "in %s", __FUNCTION__);
+
+       memset(&ev, 0, sizeof(struct iscsi_uevent));
+
+       ev.type = ISCSI_UEVENT_DESTROY_SESSION_ASYNC;
+       ev.transport_handle = transport_handle;
+       ev.u.d_session.sid = sid;
+
+       iov[1].iov_base = &ev;
+       iov[1].iov_len = sizeof(ev);
+       rc = __kipc_call(iov, 2);
+       if (rc < 0)
+               return rc;
+
+       return 0;
+}
+
 static int
 kunbind_session(uint64_t transport_handle, uint32_t sid)
 {
@@ -1763,6 +1787,7 @@ struct iscsi_ipc nl_ipc = {
        .sendtargets            = ksendtargets,
        .create_session         = kcreate_session,
        .destroy_session        = kdestroy_session,
+       .destroy_session_async  = kdestroy_session_async,
        .unbind_session         = kunbind_session,
        .create_conn            = kcreate_conn,
        .destroy_conn           = kdestroy_conn,
```

2. in session_conn_shutdown(), try destroy_session_async first:

```
diff --git a/usr/initiator.c b/usr/initiator.c
index 7186729..98b07a8 100644
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -498,15 +498,24 @@ session_conn_shutdown(iscsi_conn_t *conn, queue_task_t *qtask,
 
 cleanup:
        if (session->id != INVALID_SESSION_ID) {
+               int err = 0;
                log_debug(2, "kdestroy session %u", session->id);
                session->r_stage = R_STAGE_SESSION_DESTOYED;
-               if (ipc->destroy_session(session->t->handle, session->id)) {
-                       log_error("can not safely destroy session %d",
-                                 session->id);
-                       return ISCSI_ERR_INTERNAL;
+               err = ipc->destroy_session_async(session->t->handle, session->id);
+               if (err == -ENOSYS) {
+                       err = ipc->destroy_session(session->t->handle, session->id);
+                       if (err) {
+                               log_error("can not safely destroy session %d",
+                                       session->id);
+                               return ISCSI_ERR_INTERNAL;
+                       } else {
+                               goto out;
+                       }
                }
+               return err ? ISCSI_ERR_INTERNAL : ISCSI_SUCCESS;
        }
 
+out:
        log_warning("Connection%d:%d to [target: %s, portal: %s,%d] "
                    "through [iface: %s] is shutdown.",
                    session->id, conn->id, session->nrec.name,
```

3. in iscsi_async_session_destruction(), cleaning up session and response the qtask. But it's hard to implement the response qtask with current code(no easy way to pass qtask from session_conn_shutdown to iscsi_async_session_destruction). 
It would be easy to do this based on your changes(iscsid: simplify qtask handling) where we can figure out the qtask easily. BTY, I really like this changes, it make the qtask handle clear and easy to understand, it can also fix the issue of failed to logout an unrecoverable session once iscsid is restarted.
So when would you make your change merged?

wenchao-hao:
@mikechristie 
I found another issue which kernel would sending ISCSI_KEVENT_UNBIND_SESSION twice. It is fixed in my forth commit.
Maybe we need to change kernel's iscsi_remove_session() function to address this issue, like remove calling of __iscsi_unbind_session().

mikechristie:
> @mikechristie I found another issue which kernel would sending ISCSI_KEVENT_UNBIND_SESSION twice. It is fixed in my forth commit. Maybe we need to change kernel's iscsi_remove_session() function to address this issue, like remove calling of __iscsi_unbind_session().

It's a little complicated.

For drivers that support multiple sessions per host, we have to delete the target at some point before the session is removed and for the case where userspace doesn't do a unbind target call we need to remove the target.

It's a bug to send the event twice. We could add code to detect if an unbind call has been done. There is a check for if the target_id is still valid in __iscsi_unbind_session, but that could be hit if we unbound already or also if we never found a target, so can't just do a return instead of goto there.

The problem in the end though is supporting kernels that do the double unbind event, so we can fix the kernel but in userspace we sill need code to make sure we don't crash/fail if we get 2 events. So your approach of fixing in userspace is good. We can also fix up the kernel to make it nicer.

wenchao-hao:
> Also I should be able t review and test this next week. I got stuck on some kernel regression issues on the list this week.

OK, waiting for your review.

wenchao-hao:
Fix a bug of double free when kernel do not support async logout.

wenchao-hao:
> > @mikechristie I found another issue which kernel would sending ISCSI_KEVENT_UNBIND_SESSION twice. It is fixed in my forth commit. Maybe we need to change kernel's iscsi_remove_session() function to address this issue, like remove calling of __iscsi_unbind_session().
> 
> It's a little complicated.
> 
> For drivers that support multiple sessions per host, we have to delete the target at some point before the session is removed and for the case where userspace doesn't do a unbind target call we need to remove the target.
> 
> It's a bug to send the event twice. We could add code to detect if an unbind call has been done. There is a check for if the target_id is still valid in __iscsi_unbind_session, but that could be hit if we unbound already or also if we never found a target, so can't just do a return instead of goto there.
> 
> The problem in the end though is supporting kernels that do the double unbind event, so we can fix the kernel but in userspace we sill need code to make sure we don't crash/fail if we get 2 events. So your approach of fixing in userspace is good. We can also fix up the kernel to make it nicer.

What about do not send an ISCSI_KEVENT_UNBIND_SESSION if __iscsi_unbind_session() is called from __iscsi_destroy_session().

We can assume that ISCSI_KEVENT_UNBIND_SESSION aimed to trigger userspace to perform iscsi_stop() and session_conn_shutdown() further, while once kernel execute __iscsi_destroy_session(), the userspace has already called unbind_session or session is not in login state, which do not need kernel send ISCSI_KEVENT_UNBIND_SESSION to trigger session_conn_shutdown() flow.

I made some changes in kernel to fix this issue as following:

```
--- a/drivers/scsi/scsi_transport_iscsi.c
+++ b/drivers/scsi/scsi_transport_iscsi.c
@@ -1947,11 +1947,8 @@ void iscsi_block_session(struct iscsi_cls_session *session)
 }
 EXPORT_SYMBOL_GPL(iscsi_block_session);
 
-static void __iscsi_unbind_session(struct work_struct *work)
+static void __iscsi_unbind_session(struct iscsi_cls_session *session, int event)
 {
-       struct iscsi_cls_session *session =
-                       container_of(work, struct iscsi_cls_session,
-                                    unbind_work);
        struct Scsi_Host *shost = iscsi_session_to_shost(session);
        struct iscsi_cls_host *ihost = shost->shost_data;
        unsigned long flags;
@@ -1979,10 +1976,20 @@ static void __iscsi_unbind_session(struct work_struct *work)
                ida_simple_remove(&iscsi_sess_ida, target_id);
 
 unbind_session_exit:
-       iscsi_session_event(session, ISCSI_KEVENT_UNBIND_SESSION);
+       if (event)
+               iscsi_session_event(session, ISCSI_KEVENT_UNBIND_SESSION);
        ISCSI_DBG_TRANS_SESSION(session, "Completed target removal\n");
 }
 
+static void __iscsi_unbind_work(struct work_struct *work)
+{
+       struct iscsi_cls_session *session =
+                       container_of(work, struct iscsi_cls_session,
+                                    unbind_work);
+
+       __iscsi_unbind_session(session, 1);
+}
+
 static void __iscsi_destroy_session(struct work_struct *work)
 {
        struct iscsi_cls_session *session =
@@ -2011,7 +2018,7 @@ iscsi_alloc_session(struct Scsi_Host *shost, struct iscsi_transport *transport,
        INIT_LIST_HEAD(&session->sess_list);
        INIT_WORK(&session->unblock_work, __iscsi_unblock_session);
        INIT_WORK(&session->block_work, __iscsi_block_session);
-       INIT_WORK(&session->unbind_work, __iscsi_unbind_session);
+       INIT_WORK(&session->unbind_work, __iscsi_unbind_work);
        INIT_WORK(&session->scan_work, iscsi_scan_session);
        INIT_WORK(&session->destroy_work, __iscsi_destroy_session);
        spin_lock_init(&session->lock);
@@ -2177,7 +2184,7 @@ void iscsi_remove_session(struct iscsi_cls_session *session)
        flush_work(&session->scan_work);
        /* flush running unbind operations */
        flush_work(&session->unbind_work);
-       __iscsi_unbind_session(&session->unbind_work);
+       __iscsi_unbind_session(session, 0);
 
        /* hw iscsi may not have removed all connections from session */
        err = device_for_each_child(&session->dev, NULL,
```

usr/netlink.c
mikechristie: That sounds good. Just take my patch

https://github.com/open-iscsi/open-iscsi/pull/292/commits/0795abbadef865bf3338f127ab03c990c8690422

and integrate into your patchset so it all gets tested together.

wenchao-hao: > That sounds good. Just take my patch
> 
> [0795abb](https://github.com/open-iscsi/open-iscsi/commit/0795abbadef865bf3338f127ab03c990c8690422)
> 
> and integrate into your patchset so it all gets tested together.

I would integrate the patch and update this PR in next days.

gonzoleeman:
This has been sitting a while, with no activity. Can I close this? Then you can open a new pull request when ready?

wenchao-hao:
> This has been sitting a while, with no activity. Can I close this? Then you can open a new pull request when ready?

OK

PR #333, opened 2022-04-03
Build system tweaks from downstream

thesamesam:
Hi,

These are a few patches we've been maintaining downstream in Gentoo which should help general distribution folks out.

### DIFF
--- a/libopeniscsiusr/Makefile
+++ b/libopeniscsiusr/Makefile
@@ -89,7 +89,7 @@ install: $(LIBS) $(LIBS_MAJOR)
 	ln -sf $(LIBS) $(DESTDIR)/$(LIB_DIR)/$(LIBS_MAJOR)
 	$(INSTALL) $(HEADERS) $(DESTDIR)$(INCLUDE_DIR)/
 	$(INSTALL) -m 644 -D $(PKGFILE).in $(DESTDIR)$(PKGCONF_DIR)/$(PKGFILE)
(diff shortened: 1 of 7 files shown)


libopeniscsiusr/Makefile
```diff
@@ -108,7 +108,7 @@ docs/man/$(EXTRA_MAN_FILES).gz: $(HEADERS)
 	    perl docs/doc-preclean.pl > "$(TEMPFILE)"
 	perl docs/kernel-doc -man "$(TEMPFILE)" | \
 	    perl docs/split-man.pl docs/man
-	-rm -f "$(TEMPFILE)"
+	-$(RM) -f "$(TEMPFILE)"
```
gonzoleeman, line 111: The RM Make macro is already set to "rm -f" by default, so you need not keep the "-f". Please fix this here and in other Makefiles.

gonzoleeman (changes requested):
Except for the unneeded "-f" in "$(RM) -f", looks good. I'd be glad to merge these with that change.

thesamesam:
Yep, I'll be doing it later. Yesterday I was busy mostly IRL.

libopeniscsiusr/Makefile
thesamesam: Done, thanks!

thesamesam:
Done.

gonzoleeman (approved):

gonzoleeman:
Thank you!

thesamesam:
Thanks!

Issue #338, opened 2022-04-12
Better handle iscsi session unbind and destroy


wenchao-hao:
I found an issue which kernel would send ISCSI_KEVENT_UNBIND_SESSION twice which
shoud be fixed.

But it's a little hard to address it in an attractively elegant way.

I came up a way to fix it which is do not send ISCSI_KEVENT_UNBIND_SESSION if
__iscsi_unbind_session() is called from iscsi_remove_session(). While this way
would break other's implement if userspace did not do session unbind and expect
ISCSI_KEVENT_UNBIND_SESSION.

Another way is to introduce a flag(no matter what and which) to make sure one
session send ISCSI_KEVENT_UNBIND_SESSION. For example introduce a new session state
ISCSI_SESSION_UNBIND, check if session is in ISCSI_SESSION_UNBIND state in
__iscsi_unbind_session(), if not, then perform unbind operations and send
ISCSI_KEVENT_UNBIND_SESSION, else do not send event.

But this would break the issue fixed in following commit:
https://github.com/torvalds/linux/commit/13e60d3ba287

The issue is iscsid died for any reason after it send unbind session to kernl, and
once iscsid restarted again, it loss kernel's ISCSI_KEVENT_UNBIND_SESSION event.
Now kernel think iscsi_cls_session has already sent an ISCSI_KEVENT_UNBIND_SESSION
event and would not send any more. Which would cause userspace unable to logout. And
actually the session is in invalid state(it's target_id is INVALID)

I come up a way to address the issue above:

1. in kernel, set session to ISCSI_SESSION_UNBIND state in __iscsi_unbind_session()
   in this way, we can guarantee only one ISCSI_KEVENT_UNBIND_SESSION is sent.
2. in userspace, check session's state during iscsid restart, if session is in
   ISCSI_SESSION_UNBIND, do not sync this session and perform session teardown flow.
   I think this is reasonable, because once a session is unbinded, we can not recover
   it any more(mainly because it's target id is INVALID)

wenchao-hao:
@mikechristie Hi mike, I submit a issue and discuss the ISCSI_KEVENT_UNBIND_SESSION issue here.

wenchao-hao:
> The issue is iscsid died for any reason after it send unbind session to kernel

For this scenario, we should not try to relogin this session after iscsid restart which iscsid do now.
Following issue would happen:

1. session's id would be broken;
2. session's target id is -1
3. iscsid report session is operational after recovery

[...]

PR #339, opened 2022-04-13
Correct handle broken logout session when iscsid restarted

wenchao-hao:
Refer to  #338 for more details.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1149,6 +1149,18 @@ static void iscsi_stop(void *data)
 
 	iscsi_ev_context_put(ev_context);
 
+	/*
+	 * Once receive ISCSI_KEVENT_UNBIND_SESSION, ctldev_handle() would
+	 * trigger an EV_CONN_STOP event, this event would call iscsi_stop()
+	 * to perform connection stop operations.
+	 * When using async destroy session and older kernels we might
+	 * get more than one session unbind events. If we are already in
+	 * the IN_LOGOUT state then we know we've already got one, so we can
+	 * ignore the session unbind events
+	 */
+	if (conn->state == ISCSI_CONN_STATE_IN_LOGOUT)
+		return;
+
 	if (!(conn->session->t->caps & CAP_LOGIN_OFFLOAD)) {
 		rc = iscsi_send_logout(conn);
 		if (!rc)
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -2046,6 +2046,8 @@ sync_conn(iscsi_session_t *session, uint32_t cid)
  */
 static int check_and_cleanup_session(iscsi_session_t *session)
 {
+	char id[NAME_SIZE];
+	char state[10];
 	int rc = 0;
 
 	/*
@@ -2063,7 +2065,26 @@ static int check_and_cleanup_session(iscsi_session_t *session)
 		goto destroy_session;
 	}
 
+	snprintf(id, sizeof(id), "connection%d:0", session->id);
+	sysfs_get_str(id, "iscsi_connection", "state", state, 10);
+
+	/*
+	 * A connection would be in down state in next scenario:
+	 *
+	 * - iscsid died after stop_conn() is called during logout
+	 * - iscsid died after create_conn() is called during login
+	 */
+	if (strcmp(state, "down") == 0) {
+		log_warning("destroying session%d with down connection", session->id);
+		goto destroy_conn;
+	}
+
 	return ISCSI_SESSION_NEED_REOPEN;
+
+destroy_conn:
+	rc = ipc->destroy_conn(session->t->handle, session->id, 0);
+	if (rc)
+		log_error("BUG: clean connection%d:0 failed %s", session->id, strerror(-rc));
 
 destroy_session:
 	rc = ipc->destroy_session(session->t->handle, session->id);
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -55,6 +55,7 @@
 
 #define ISCSI_SESSION_NEED_REOPEN	0
 #define ISCSI_SESSION_DESTROYED		1
+#define ISCSI_SESSION_WAIT_UNBIND_EVENT	2
 
 #define PROC_DIR "/proc"
 
@@ -1069,7 +1070,16 @@ setup_full_feature_phase(iscsi_conn_t *conn)
 			    session->nrec.conn[conn->id].port,
 			    session->nrec.iface.name);
 	} else {
+		char state[12];
+		char id[NAME_SIZE];
+		int err;
+
 		session->notify_qtask = NULL;
+
+		snprintf(id, sizeof(id), "session%d", session->id);
+		err = sysfs_get_str(id, "iscsi_session", "target_state", state, 12);
+		if (!err && !strncmp(state, "ALLOCATED", 9))
+			session_scan_host(session, session->hostno, NULL);
 
 		session_online_devs(session->hostno, session->id);
 		mgmt_ipc_write_rsp(c->qtask, ISCSI_SUCCESS);
@@ -2043,6 +2053,7 @@ sync_conn(iscsi_session_t *session, uint32_t cid)
  * RETURN VALUE:
  * 	ISCSI_SESSION_NEED_REOPEN: session is valid and ready to reopen
  * 	ISCSI_SESSION_DESTROYED: session is kind of broken which has been destroyed
+ * 	ISCSI_SESSION_WAIT_UNBIND_EVENT: waiting for kernel's unbind event
  */
 static int check_and_cleanup_session(iscsi_session_t *session)
 {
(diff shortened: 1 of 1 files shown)


wenchao-hao:
I tested this with some changes(stop iscsid after session_unbind succeed) in iscsid and kernel with patches:

https://lore.kernel.org/linux-scsi/20220414014947.4168447-1-haowenchao@huawei.com/

Following is a brief log of test:

If session is in UNBOUND state, it would be shutdown in iscsi_sync_session.

```
root@fedora ~:# systemctl status iscsid 
× iscsid.service - Open-iSCSI
     Loaded: loaded (/usr/lib/systemd/system/iscsid.service; disabled; vendor preset: disabled)
     Active: failed (Result: signal) since Thu 2022-04-14 04:28:10 EDT; 4min 12s ago
TriggeredBy: ● iscsid.socket
       Docs: man:iscsid(8)
             man:iscsiuio(8)
             man:iscsiadm(8)
    Process: 10574 ExecStart=/sbin/iscsid -f (code=killed, signal=KILL)
   Main PID: 10574 (code=killed, signal=KILL)
     Status: "Ready to process requests"
        CPU: 45ms

Apr 14 04:28:04 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:05 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:06 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:07 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:08 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:09 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:10 fedora iscsid[10574]: iscsid: sleep with session_unbind succeed
Apr 14 04:28:10 fedora systemd[1]: iscsid.service: Main process exited, code=killed, status=9/KILL
Apr 14 04:28:10 fedora systemd[1]: iscsid.service: Failed with result 'signal'.
Apr 14 04:28:10 fedora systemd[1]: Stopped Open-iSCSI.
root@fedora ~:# systemctl start iscsid 
root@fedora ~:# systemctl status iscsid 
● iscsid.service - Open-iSCSI
     Loaded: loaded (/usr/lib/systemd/system/iscsid.service; disabled; vendor preset: disabled)
     Active: active (running) since Thu 2022-04-14 04:32:33 EDT; 1s ago
TriggeredBy: ● iscsid.socket
       Docs: man:iscsid(8)
             man:iscsiuio(8)
             man:iscsiadm(8)
   Main PID: 10659 (iscsid)
     Status: "Ready to process requests"
      Tasks: 1 (limit: 38062)
     Memory: 1.8M
        CPU: 21ms
     CGroup: /system.slice/iscsid.service
             └─10659 /sbin/iscsid -f

Apr 14 04:32:32 fedora systemd[1]: Starting Open-iSCSI...
Apr 14 04:32:32 fedora iscsid[10659]: iscsid: Shutdown unbound session3
Apr 14 04:32:32 fedora iscsid[10659]: iscsid: Connection3:0 to [target: iqn.2003-01.org.linux-iscsi.localhost.test:sn.12345678abcd, portal: 9.84.111.6,3260] through [iface: default] is shutdown.
Apr 14 04:32:33 fedora systemd[1]: Started Open-iSCSI.
root@fedora ~:# iscsiadm -m session
iscsiadm: No active sessions.
root@fedora ~:# ls /sys/class/iscsi_session/
```

usr/iscsid.c
```diff
@@ -202,11 +202,19 @@ static int sync_session(__attribute__((unused))void *data,
...
 
 	log_debug(7, "sync session [%d][%s,%s.%d][%s]", info->sid,
 		  info->targetname, info->persistent_address,
 		  info->port, info->iface.hwaddress);
 
+	iscsi_sysfs_get_session_state(session_state, info->sid);
+
+	if (!strcmp(session_state, "FREE")) {
+		log_error("Skip syncing a freed session%d", info->sid);
+		return 0;
```
mikechristie, line 215: The FREE state doesn't always mean we are freeing the session during removal. It means the session is no longer valid. This could be because it is being removed or the replacement/recovery timeout fired (see session_recovery_timedout in scsi_transport_iscsi.c). For the latter, if we re-login we will re-use the session struct. So if iscsid was restarted for the recovery timed out case we would still want to try to relogin/sync when iscsid starts up again.

I think you need to check if there is also a connection in sysfs (/sys/class/iscsi_connection). If there is no connection, then we were in FREE because it's being removed.

The problem above is that we've looped over sysfs and gathered the session and connection data, so we would expect that we would only get to this point if there was a valid connection. However, iscsi_sysfs_get_sessioninfo_by_id looks like it has a bug where it assumes that if we have a session then there will always be a connection, so if those

sysfs_get_str(ISCSI_CONN_SUBSYS)

calls fail, we just ignore it more or less so we can end up running sync_session.

So  I think we need 3 things:

1. In initiator.c iscsi_sync_session we could add some code that can check if there is a connection and session. If there is only a session and it's in FREE, then do like you did above and just return because it's being freed in the kernel.

2. If there is a connection and it's state is DOWN, then something was removing the session/connection but iscsid failed when we were removing the connection. Don't sync. Just finish the removal. If the conn state is DOWN then stop conn with STOP_CONN_TERM was already called. So we just need to do destroy_conn then destroy_session.

3. Fix up iscsi_sysfs_get_sessioninfo_by_id so it handles the case where the connection has been removed but the session is still there. Since for the above case, we want iscsid to cleanup we can't just return error and ignore the session. I'm not sure what we want to do here. Maybe just make sure the address/port are set to invalid values so the caller can figure out what to do?

usr/initiator.c
```diff
@@ -2050,12 +2051,19 @@ iscsi_sync_session(node_rec_t *rec, queue_task_t *qtask, uint32_t sid)
...
-
 	err = sync_conn(session, 0);
 	if (err)
 		goto destroy_session;
 
+	iscsi_sysfs_get_session_target_unbound(&target_unbound, sid);
+
+	if (target_unbound) {
+		log_error("Shutdown unbound session%d", sid);
+		session_conn_shutdown(&session->conn[0], qtask, 0);
```
mikechristie, line 2062: The conn->state will be ISCSI_CONN_STATE_CLEANUP_WAIT from sync_conn() right?

If so, the conn could have been in the logged in state when the unbind was running and iscsid stopped. So we still need to do a stop_conn call so the kernel can run libiscsi.c:iscsi_conn_stop.

I think we can check the conn kernel state to detect if it's in the DOWN state already (scsi_transport_iscsi.c:iscsi_stop_conn sets that state when userspace has called into the kernel to stop a conn for removal).

wenchao-hao: > The conn->state will be ISCSI_CONN_STATE_CLEANUP_WAIT from sync_conn() right?
> 

Yes, it's now ISCSI_CONN_STATE_CLEANUP_WAIT state.

> If so, the conn could have been in the logged in state when the unbind was running and iscsid stopped. So we still need to do a stop_conn call so the kernel can run libiscsi.c:iscsi_conn_stop.
> 

Yes, we need to check it and judge if stop_conn is necessary now.

> I think we can check the conn kernel state to detect if it's in the DOWN state already (scsi_transport_iscsi.c:iscsi_stop_conn sets that state when userspace has called into the kernel to stop a conn for removal).

gonzoleeman:
What's the status of this pull request? Sorry, I didn't take time to re-read the whole thread, but now that Mike has submitted a few changes upstream, is this still an issue?

mikechristie:
It's not related to any patches upstream so it's still needed.

wenchao-hao:
> What's the status of this pull request? Sorry, I didn't take time to re-read the whole thread, but now that Mike has submitted a few changes upstream, is this still an issue?

Apolgy for my long time absence. I was too busy to submit changes in last 2 month. It seems I would have time to make it in next month.

gonzoleeman:
> > What's the status of this pull request? Sorry, I didn't take time to re-read the whole thread, but now that Mike has submitted a few changes upstream, is this still an issue?
> 
> Apolgy for my long time absence. I was too busy to submit changes in last 2 month. It seems I would have time to make it in next month.

No problem at all. I understand being busy. I just wanted to clean this request up if it was no longer needed.

wenchao-hao:
@mikechristie @gonzoleeman  Hi mike and lee, I found some other scenario which may cause iscsi session in invalid state, and I am trying to figure these out. This PR would be updated in next few days

wenchao-hao:
> @mikechristie @gonzoleeman  Hi mike and lee, I found some other scenario which may cause iscsi session in invalid state, and I am trying to figure these out. This PR would be updated in next few days

Updated

violethaze74 (approved):

wenchao-hao:
Update this PR with kernel patch:https://lore.kernel.org/all/20220802034729.2566787-1-haowenchao@huawei.com/

For kernel which did not apply the target_state patch, when iscsid restart, it would show:

```
root@fedora /code/open-iscsi:# ./usr/iscsid -f
iscsid: failed to open session's target state No such file or directory
iscsid: connection1:0 is operational after recovery (1 attempts)
```

For kernel which applied the target_state patch, when iscsid restart:

- If the target_state is UNBINDING, it would show:

```
root@fedora /code/open-iscsi:# ./usr/iscsid -f
Donot sync UNBINDING session3
iscsid: connection3:0 Could not send logout pdu(Invalid argument) from iscsi_stop. Dropping session
iscsid: Connection3:0 to [target: iqn.1994-05.com.redhat:e1d3c4cb3d4, portal: 192.168.1.12,3260] through [iface: default] is shutdown.
```

- If the target_state is UNBOUND, it would show:

```
iscsid: Shutdown UNBOUND session1
iscsid: Connection1:0 to [target: iqn.1994-05.com.redhat:e1d3c4cb3d4, portal: 192.168.1.12,3260] through [iface: default] is shutdown.
```

mikechristie:
Hey wenchao-hao,

Thanks for your work on this. I'm on vacation for a couple weeks. I'll be able to look at this and the kernel patch more when I get back.

gonzoleeman:
It looks like there are still 3 "unresolved conversations" here. I've set up open-iscsi so that they must be resolved before a merge is allowed.

@mikechristie are your issues resolved?

wenchao-hao:
> It looks like there are still 3 "unresolved conversations" here. I've set up open-iscsi so that they must be resolved before a merge is allowed.
> 
> @mikechristie are your issues resolved?

The last 3 commits are based a kernel patch, I have send the patch to maillist:
https://lore.kernel.org/linux-scsi/20221126010752.231917-1-haowenchao@huawei.com/

I think this PR could be merged after that patch is applied.

wenchao-hao:
I would submit a new PR in next days, so close this one.

PR #340, opened 2022-04-13
Make fwparam_ibft build more reliable.

gonzoleeman:
This build has lex and bison source files, but
they are not normally used to regenerate C files,
since that has been done ahead of time, statically,
to handle the case when users building the software
may not have these tools. But some build systems
think the generated C files need to be regenerated
because the timestamp on the source and generated
files is so close, just because of the order the
files were populated in this directory. So we tell
the Makefile to only generate C files if those
files are missing. This should have no effect,
since these files should never need to be
regenerated unless we change lex or bison logic.

### DIFF
--- a/usr/fwparam_ibft/Makefile
+++ b/usr/fwparam_ibft/Makefile
@@ -24,6 +24,8 @@ ifeq ($(TOPDIR),)
 	TOPDIR = ../..
 endif
 
+BISON = bison
+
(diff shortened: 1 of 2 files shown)


gonzoleeman:
This fixes an issue with my recent build re-org that seems to call "lex" when unneeded. The build is set up so that lex and bison should never need to be called.

Issue #337, opened 2022-04-11
Failed to link iscsiuio during build


wenchao-hao:
commit 9fbd6009cd917f1152a367fa7e5ae3993133c1e4 changed the Makefile.am
It's ok with fedora, while for my old version distro the building failed. It seems iscsiuio_LIBS did not work

```
ps.c' || echo './'`../../../utils/sysdeps/sysdeps.c
mv -f .deps/iscsiuio-packet.Tpo .deps/iscsiuio-packet.Po
mv -f .deps/iscsiuio-main.Tpo .deps/iscsiuio-main.Po
mv -f .deps/iscsiuio-sysdeps.Tpo .deps/iscsiuio-sysdeps.Po
mv -f .deps/iscsiuio-nic_nl.Tpo .deps/iscsiuio-nic_nl.Po
mv -f .deps/iscsiuio-ping.Tpo .deps/iscsiuio-ping.Po
mv -f .deps/iscsiuio-nic.Tpo .deps/iscsiuio-nic.Po
mv -f .deps/iscsiuio-nic_utils.Tpo .deps/iscsiuio-nic_utils.Po
mv -f .deps/iscsiuio-iscsid_ipc.Tpo .deps/iscsiuio-iscsid_ipc.Po
/bin/sh ../../libtool  --tag=CC   --mode=link gcc -I../../src/uip -I../../src/apps/brcm-iscsi -I../../src/apps/dhcpc -I../../src/unix/libs -I../../../include -I../../../usr  -DBYTE_ORDER=LITTLE -g -O2 -Wall   -o iscsiuio iscsiuio-build_date.o iscsiuio-main.o iscsiuio-clock-arch.o iscsiuio-logger.o iscsiuio-nic.o iscsiuio-nic_id.o iscsiuio-nic_vlan.o iscsiuio-nic_nl.o iscsiuio-nic_utils.o iscsiuio-packet.o iscsiuio-iscsid_ipc.o iscsiuio-ping.o iscsiuio-sysdeps.o  ../../src/uip/lib_iscsi_uip.a ../../src/apps/dhcpc/lib_apps_dhcpc.a ../../src/apps/brcm-iscsi/lib_apps_brcm_iscsi.a ../../src/unix/libs/lib_iscsiuio_hw_cnic.a -lsystemd 
[...]

PR #341, opened 2022-04-14
iscsiuio: fix LDADD

thesamesam:
iscsiuio: fix LDADD

- We don't want to use _LIBS here, we're building a library, not an executable.
- We want LDADD for the whole lot (not a mix of LDFLAGS/LIBS/LDADD),
but put the objects/libraries (.a files) first so that -l* works correctly.

Closes: https://github.com/open-iscsi/open-iscsi/issues/337
Fixes: 9fbd6009cd917f1152a367fa7e5ae3993133c1e4
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/iscsiuio/src/unix/Makefile.am
+++ b/iscsiuio/src/unix/Makefile.am
@@ -27,16 +27,13 @@ iscsiuio_CFLAGS = 	$(AM_CFLAGS)		\
 			$(LIBNL_CFLAGS)		\
 			-DBYTE_ORDER=@ENDIAN@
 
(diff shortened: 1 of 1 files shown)


wenchao-hao:
Maybe you can explain how this change can fix the issue https://github.com/open-iscsi/open-iscsi/issues/337 in commit message, I don't know if others can understand it.

thesamesam:
Sure, I'll touch up the commit message later today. Thanks for confirming it works.

gonzoleeman:
Looks good to me. Thank you.

PR #343, opened 2022-04-21
iscsid: fix logout pdu send failure detection handling

mikechristie:
This fixes a bug where when we send logout pdu we are incorrectly
handling it as an error. This was added in

https://github.com/open-iscsi/open-iscsi/pull/317

I think there might have been a mistake for which version of that patch
got merged because there are some unhandled review comments in the merged
code.

The patch also fixes some review comments that were given in that review
but were not handled when it was merged:
1. No space before "Dropping session"
2. Don't split error messages.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1153,8 +1153,9 @@ static void iscsi_stop(void *data)
 		rc = iscsi_send_logout(conn);
 		if (!rc)
 			return;
-		conn_error(conn, "Could not send logout pdu(%s) from iscsi_stop."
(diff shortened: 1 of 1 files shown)


wenchao-hao:
> This fixes a bug where when we send logout pdu we are incorrectly handling it as an error. This was added in
> 
> #317 

Sorry for that bug, and thank you for fixing it.

gonzoleeman (approved):

gonzoleeman:
Thanks Mike.

PR #347, opened 2022-05-09
Fix iscsiuio libpthread build issue

gonzoleeman:
Commit 9fbd6009cd91 ("iscsiuio: respect LDFLAGS fully")
changed iscsiuio_LDFLAGS in iscsiuio/src/unix/Makefile.am
to iscsiuio_LIBS. This means the targets listed
here move to the front of the list of arguments to
the linkage command. This causes error messages during
builds (on some architectures), such as:

> /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: iscsiuio-iscsid_ipc.o: undefined reference to symbol 'pthread_join@@GLIBC_2.2.5'

This is because the object file referencing pthread_join
is listed on the linkage command line *after* libpthread
is listed.

So move libpthread (and libdl) to iscsiuio_LADD, so that
they show up after the object files on the linkage
comand line.

Tested on x86 64-bit on openSUSE Leap 15.3 and Tumbleweed.

Fixes: 9fbd6009cd917f1152a367fa7e5ae3993133c1e4

### DIFF
--- a/iscsiuio/src/unix/Makefile.am
+++ b/iscsiuio/src/unix/Makefile.am
@@ -28,15 +28,15 @@ iscsiuio_CFLAGS = 	$(AM_CFLAGS)		\
 			-DBYTE_ORDER=@ENDIAN@
 
 iscsiuio_LIBS = 	$(AM_LIBS)		\
-			-ldl			\
 			-rdynamic		\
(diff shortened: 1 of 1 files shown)


gonzoleeman:
Resetting

PR #350, opened 2022-05-16
Add ability for MGMT IPC to check UID only

ericwmackay:
The default behavior in iscsid, which is to check mgmt IPCs
for UID==0 and user explicitly named "root", is unchanged.

This option to perform only the UID check for management
IPCs can be enabled in iscsid.conf. This can be useful for
running iscsid in a constrained environment, e.g., initramfs.

For instance, klibc initramfs in Ubuntu and Debian does not
include a user DB. Attempting to run iscsid and iscsiadm
in klibc initramfs would result in ISCSI_ERR_ACCESS for
all MGMT IPCs. Specifying just the UID check allows iscsid
and iscsiadm to run in klibc initramfs without a user DB.
systemd can still start an iscsid process later in boot with
the default full MGMT IPC auth checks, leaving the steady-state
behavior of iscsid the same as before.

The following setting in iscsid.conf enables this behavior:
iscsid.ipc_auth_uid = Yes

iscsistart now uses this mechanism to only perform the UID
check, rather than statically linking against an alternate
implementation of getpwuid(). This is effectively the same
behavior as before.


mikechristie:
Looks good to me.

gonzoleeman:
Thank you.

PR #351, opened 2022-05-19
Use config for iscsistart and iscsiadm fw login

ericwmackay:
Specifying name-value pairs as arguments to iscsistart or
iscsiadm can become unwieldy very quickly, and is less flexible
than using a config file. If a user desires to update settings,
modifying a config file and rebuilding initramfs can be simpler
than modifying arguments directly in initramfs code or scripts.

Node records created from boot context are populated with
defaults, then any settings specified in a config file are
applied (if config is present).

FW and user-specified params are still applicable. User-specified
name-value arguments are applied after config settings, so they
can serve as a safeguard against config file changes if
desired.

Added -c|--config options for iscsistart to specify the config

Updated iscsistart man page

### DIFF
--- a/usr/iscsistart.c
+++ b/usr/iscsistart.c
@@ -61,11 +61,13 @@ static LIST_HEAD(targets);
 static LIST_HEAD(user_params);
 
 static char program_name[] = "iscsistart";
+static char config_file[TARGET_NAME_MAXLEN];
 
 /* used by initiator */
 extern struct iscsi_ipc *ipc;
 
 static struct option const long_options[] = {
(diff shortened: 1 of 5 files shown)


mikechristie:
Looks ok to me.

usr/iscsistart.c
```diff
@@ -270,6 +273,11 @@ static int login_session(struct node_rec *rec)
 	return rc;
 }
 
+static char *get_config_file(void)
```
gonzoleeman, line 276: Where is this used?

gonzoleeman (approved):

PR #354, opened 2022-07-07
etc: install system unit with without executable bit

eworm-de:
All systemd unit files are expected to be not executable, so install
with correct permissions.


gonzoleeman (approved):

gonzoleeman:
Looks good. Thank you! I just fixed the exact same thing for my distro by updating the SPEC file, but for "make install" this fixes it too.

PR #357, opened 2022-07-29
make: avoid hard-coding path to sed

zeha:
Just use PATH to find sed. Not all distributions have moved to a /usr-merged
layout yet.

For Debian I had to patch this, as we still need to build on layouts where sed is (only) in /bin/sed. I think this should work just fine on Fedora, SUSE, etc.


gonzoleeman:
Thank you!

Issue #38, opened 2017-01-25
iscsiadm


royrosen:
Hi,
I am using iscsiadm to create many connections.
I can see that isid fields a and b contain the driver id, isid c has a number 0-0xFF but isid d is always 0.
so if I create more than 256 logins on the same session new logins get the same isid since only field c is changing and it has only 1 byte.
Can you please tell if this is misconfiguration, because from the code it seems that the id should be written into fields c and d together, so why is the always in the range of 0-0xFF?
Thanks, Roy.

royrosen:
For example, running
iscsiadm -m session -r /sys/devices/platform/host891/session883/ -o new
257 times would mean that the first and last login are getting the same isid.
Can you please tell what is wrong here?

cleech:
This should be fixed with commit 4c1f2d9
I responded on the mailing list regarding distro packages that include this change

PR #358, opened 2022-08-01
Change the calling order of ep_disconnect() and stop_conn()

windfine:
I got a kernel NULL pointer derference report as below:

    BUG: kernel NULL pointer dereference, address: 0000000000000038
    #PF: supervisor read access in kernel mode
    #PF: error_code(0x0000) - not-present page
    PGD 0 P4D 0
    Oops: 0000 [#1] PREEMPT SMP NOPTI
    CPU: 4 PID: 1050 Comm: cat Not tainted 5.19.0 #38
    RIP: 0010:kernel_getpeername+0xf/0x30
    Call Trace:
     <TASK>
     ? iscsi_sw_tcp_conn_get_param+0x11f/0x17f
     show_conn_ep_param_ISCSI_PARAM_CONN_ADDRESS+0x90/0xb0
     dev_attr_show+0x1d/0x50
     sysfs_kf_seq_show+0xad/0x120
     kernfs_seq_show+0x2c/0x40
     seq_read_iter+0x12e/0x4d0
     ? aa_file_perm+0x177/0x590
     kernfs_fop_read_iter+0x183/0x210
     new_sync_read+0xfe/0x180
     ? 0xffffffff81000000
     vfs_read+0x14d/0x1a0
     ksys_read+0x6d/0xf0
     __x64_sys_read+0x1a/0x20
     do_syscall_64+0x3b/0x90
     entry_SYSCALL_64_after_hwframe+0x63/0xcd

The life cycle of socket is:
login:  ep_connect(create sock)   ---> bind_conn(bind sock to conn)
logout: ep_disconnect(close sock) ---> stop_conn(unbind sock from conn)
The calling order is wrong on logout, will cause this issue in the
following scenarios:

             iscsi logout               get sock param by sysfs
    --------------------------------  ---------------------------
        User Space | Kernel Space             Kernel Space
    --------------------------------  ---------------------------
    session_conn_shutdown
      ep_disconnect
        close(conn->socket_fd)
                     sock_close
                       __sock_release
                         sock->ops = NULL

                                     iscsi_sw_tcp_conn_get_param
                                       sock = tcp_sw_conn->sock;
                                         kernel_getsockname
                                           sock->ops->getname
                                           ^^^^^^^^^
                                           NULL pointer dereference

     ipc->stop_conn
       __kipc_call
                    iscsi_sw_tcp_conn_stop
                      iscsi_sw_tcp_release_conn
                        sock = NULL

So change the calling order of ep_disconnect() and stop_conn() to
fix this issue.

Signed-off-by: Li Jinlin <lijinlin3@huawei.com>

### DIFF
--- a/usr/discovery.c
+++ b/usr/discovery.c
@@ -1032,9 +1032,6 @@ static void iscsi_destroy_session(struct iscsi_session *session)
 		goto done;
 	}
 
-	log_debug(2, "%s ep disconnect", __FUNCTION__);
-	t->template->ep_disconnect(conn);
-
 	log_debug(2, "stop conn");
 	rc = ipc->stop_conn(session->t->handle, session->id,
 			   conn->id, STOP_CONN_TERM);
@@ -1043,6 +1040,9 @@ static void iscsi_destroy_session(struct iscsi_session *session)
 			  session->id, conn->id, rc);
 		goto done;
         }
+
+	log_debug(2, "%s ep disconnect", __FUNCTION__);
+	t->template->ep_disconnect(conn);
 
 	log_debug(2, "%s destroy conn", __FUNCTION__);
         rc = ipc->destroy_conn(session->t->handle, session->id, conn->id);
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -467,11 +467,6 @@ session_conn_shutdown(iscsi_conn_t *conn, queue_task_t *qtask,
 {
 	iscsi_session_t *session = conn->session;
 
-	conn_debug(2, conn, "disconnect conn");
-	/* this will check for a valid interconnect connection */
-	if (session->t->template->ep_disconnect)
-		session->t->template->ep_disconnect(conn);
-
 	if (session->id == INVALID_SESSION_ID)
 		goto cleanup;
 
@@ -488,6 +483,11 @@ session_conn_shutdown(iscsi_conn_t *conn, queue_task_t *qtask,
 			return ISCSI_ERR_INTERNAL;
 		}
 	}
+
+	conn_debug(2, conn, "disconnect conn");
+	/* this will check for a valid interconnect connection */
+	if (session->t->template->ep_disconnect)
+		session->t->template->ep_disconnect(conn);
 
 	conn_debug(2, conn, "kdestroy conn");
 	if (ipc->destroy_conn(session->t->handle, session->id,
@@ -587,7 +587,6 @@ __session_conn_reopen(iscsi_conn_t *conn, queue_task_t *qtask, int do_stop,
 	conn_delete_timers(conn);
 	conn->state = ISCSI_CONN_STATE_XPT_WAIT;
 
-	conn->session->t->template->ep_disconnect(conn);
 	if (do_stop) {
 		/* state: ISCSI_CONN_STATE_CLEANUP_WAIT */
 		if (ipc->stop_conn(session->t->handle, session->id,
@@ -600,6 +599,8 @@ __session_conn_reopen(iscsi_conn_t *conn, queue_task_t *qtask, int do_stop,
 		log_debug(3, "connection %d:%d is stopped for recovery",
 			  session->id, conn->id);
 	}
+
+	conn->session->t->template->ep_disconnect(conn);
 
 	if (!redirected) {
 		delay = session->def_time2wait;


mikechristie:
Hi,

3 comments:

1. Yeah, the bind, stop, start and ep_connect/disconnect are a little weird because:

ep_connect goes with ep_disconnect
conn_start goes with conn_stop
bind has no matching call.

We do the disconnect before first before stop for offload drivers. For offload, the userspace ep_disconnect call will end up calling into the offload kernel driver's ep_disconnect callout and after that returns the drivers will not be executing IO. We can then do stop_conn which will clean up running IO in the kernel knowing that the driver is not accessing the cmd's memory anymore.

If you call stop_conn first then that calls into the kernel and offload driver which will cleanup running commands in a way that memory could be freed while the driver is still executing commands and we have a use after free.

So we will need to take that into account in the fix.

2. For software iscsi and this bug, are you sure the analysis is correct? I mean are you sure

ep_disconnect -> close(conn->socket_fd) will do sock_close and call __sock_release at this time?

Doesn't close() just do a put on the socket's fd?

Then is the last put done in iscsi_sw_tcp_conn_stop -> iscsi_sw_tcp_release_conn -> sockfd_put? So sock->ops should be set until the last sockfd_put by the conn_stop path.

If we did do __sock_release when ep_disconnect is called then we should hit a NULL sock->ops ref when iscsi_sw_tcp_conn_stop -> iscsi_sw_tcp_release_conn -> kernel_sock_shutdown and that does sock->ops->shutdown.

3. Could there be a race where we have called 

1. iscsi_sw_tcp_conn_get_param takes frwd_lock and does sock_hold() then drops the frwd_lock. sock_hold does a get on the "struct sock".
2. iscsi_sw_tcp_release_conn does  sockfd_put() which does the last put on the "struct socket". This now does the last put and calls __sock_release.
3.  iscsi_sw_tcp_conn_get_param calls kernel_getsockname and access the NULL sock->ops.

the problem here is that iscsi_sw_tcp_conn_get_param did a get on the "struct sock", but we needed a get on the "struct socket".

mikechristie:
We could re-write the interface to have a unbind and do things in an order that makes sense, but it would be difficult for backwards compat.

I think for just fixing this issue quickly, iscsi_sw_tcp_conn_get_param could do a fget(sock) and sockfd_put(sock);

windfine:
> Hi,
> 
> 3 comments:
> 
> 1. Yeah, the bind, stop, start and ep_connect/disconnect are a little weird because:
> 
> ep_connect goes with ep_disconnect conn_start goes with conn_stop bind has no matching call.
> 
> We do the disconnect before first before stop for offload drivers. For offload, the userspace ep_disconnect call will end up calling into the offload kernel driver's ep_disconnect callout and after that returns the drivers will not be executing IO. We can then do stop_conn which will clean up running IO in the kernel knowing that the driver is not accessing the cmd's memory anymore.
> 
> If you call stop_conn first then that calls into the kernel and offload driver which will cleanup running commands in a way that memory could be freed while the driver is still executing commands and we have a use after free.
> 
> So we will need to take that into account in the fix.
> 

Thank you for the explanation

> 2. For software iscsi and this bug, are you sure the analysis is correct? I mean are you sure
> 
> ep_disconnect -> close(conn->socket_fd) will do sock_close and call __sock_release at this time?
> 
> Doesn't close() just do a put on the socket's fd?
> 
> Then is the last put done in iscsi_sw_tcp_conn_stop -> iscsi_sw_tcp_release_conn -> sockfd_put? So sock->ops should be set until the last sockfd_put by the conn_stop path.
> 
> If we did do __sock_release when ep_disconnect is called then we should hit a NULL sock->ops ref when iscsi_sw_tcp_conn_stop -> iscsi_sw_tcp_release_conn -> kernel_sock_shutdown and that does sock->ops->shutdown.
> 

 The analysis is uncorrect, close(conn->socket_fd) does fput is not the last put, won't  do sock_close.

> 3. Could there be a race where we have called
> 
> 3.1. iscsi_sw_tcp_conn_get_param takes frwd_lock and does sock_hold() then drops the frwd_lock. sock_hold does a get on the "struct sock". 3.2. iscsi_sw_tcp_release_conn does sockfd_put() which does the last put on the "struct socket". This now does the last put and calls __sock_release. 3.3. iscsi_sw_tcp_conn_get_param calls kernel_getsockname and access the NULL sock->ops.
> 
> the problem here is that iscsi_sw_tcp_conn_get_param did a get on the "struct sock", but we needed a get on the "struct socket".

Yes, this is the coorect ananysis.

windfine:
> We could re-write the interface to have a unbind and do things in an order that makes sense, but it would be difficult for backwards compat.
> 
> I think for just fixing this issue quickly, iscsi_sw_tcp_conn_get_param could do a fget(sock) and sockfd_put(sock);

I will try to push new patch according this, thanks.

PR #361, opened 2022-08-21
Shellcheck-clean (and debashify where applicable) all shell programs

nabijaczleweli:
This removes all bash uses and fixes all shellcheck warnings.

Also a manual clarification.

### DIFF
--- a/doc/iscsi_fw_login.8
+++ b/doc/iscsi_fw_login.8
@@ -4,9 +4,9 @@ iscsi_fw_login \- Login to all iSCSI Firmware targets
 .SH SYNOPSIS
 .B iscsi_fw_login
 .SH DESCRIPTION
-This helper function logs into all known iSCSI firmware targets.
+Logs into all known iSCSI firmware targets.
 .P
-It is meant as a helper function to be called by udev when new
+Meant as an udev hook to be called by udev when new
 firmware targets are detected asynchronously.
 .SH AUTHOR
 Written by Lee Duncan
--- a/etc/systemd/iscsi-init.service.template
+++ b/etc/systemd/iscsi-init.service.template
@@ -16,4 +16,4 @@ WantedBy=systemd-remount-fs.service
 [Service]
 Type=oneshot
 RemainAfterExit=no
-ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`@SBINDIR@/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
+ExecStart=/bin/sh -c 'echo "InitiatorName=`@SBINDIR@/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
--- a/etc/initd/initd.debian
+++ b/etc/initd/initd.debian
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/bin/sh
 ### BEGIN INIT INFO
 # Provides:
 # Required-Start:
@@ -21,8 +21,6 @@ if [ ! -d /sys/class/ ]; then
   log_failure_msg "iSCSI requires a mounted sysfs, not started."
   exit 1
 fi
-
-nodestartup_re='s/^node\.conn\[0]\.startup[ 	]*=[ 	]*//p'
 
 RETVAL=0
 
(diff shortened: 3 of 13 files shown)


gonzoleeman:
What is the reason for changing bash to sh? It seems like it requires some changes to the scripts, which in turn adds a chance for some errors to occur.

And what was your testing, i.e. how did you test these changes (since there is only one test suite, and I'm pretty sure it doesn't test many of the scripts)?

doc/iscsi_fw_login.8
```diff
@@ -4,9 +4,9 @@ iscsi_fw_login \- Login to all iSCSI Firmware targets
 .SH SYNOPSIS
 .B iscsi_fw_login
 .SH DESCRIPTION
-This helper function logs into all known iSCSI firmware targets.
+Logs into all known iSCSI firmware targets.
 .P
-It is meant as a helper function to be called by udev when new
+Meant as an udev hook to be called by udev when new
```
gonzoleeman, line 9: I believe that should be "Meant as a udev hook ...".

etc/systemd/iscsi-init.service.template
```diff
@@ -16,4 +16,4 @@ WantedBy=systemd-remount-fs.service
 [Service]
 Type=oneshot
 RemainAfterExit=no
-ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`@SBINDIR@/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
+ExecStart=/bin/sh -c 'echo "InitiatorName=`@SBINDIR@/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
```
gonzoleeman, line 19: Why do you think "/bin/sh" in the *only* valid name? And you misspelled "name" in your commit message.

doc/iscsi_fw_login.8
nabijaczleweli: It's /ɤdəv/ in my accent, so an; you might americanise this to /judəv/ i guess, which would be a?

etc/systemd/iscsi-init.service.template
nabijaczleweli: If you want a citation, then https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s04.html or https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s04.html#specificOptions11 (alas, it's FHS, oh well).

What's harder to cite is that that's because (barring the anomaly that was SVr4 in particular with `/sbin/sh`), since V1, that's *the* location of the shell.

/usr/bin/sh is only valid on merged-/usr systems (in which case /bin *is* /usr/bin). Trivially, not all systems are merged-/usr.

nabijaczleweli:
Conversely, I ask you this: what is the reason to require bash when none of the programs actually use any bash extensions (beyond, like, abusing `==`, overusing `=~`, a `&>` and a `$EUID`)? The actual bash removal parts are maybe 20 lines (excluding shebangs) of this; the vast majority is glaring holes found by shellcheck. Most transformations are trivial, the ones that aren't I tested in situ if possible or ex situ if I lacked the configuration for it. For checking any booboos I defer to your capable eyes and the test-suite.

doc/iscsi_fw_login.8
gonzoleeman: It's a nit, but the correct grammar in english is to use "a".

etc/systemd/iscsi-init.service.template
gonzoleeman: Okay, I'll accept that then.

gonzoleeman:
> Conversely, I ask you this: what is the reason to require bash when none of the programs actually use any bash extensions (beyond, like, abusing `==`, overusing `=~`, a `&>` and a `$EUID`)? The actual bash removal parts are maybe 20 lines (excluding shebangs) of this; the vast majority is glaring holes found by shellcheck. Most transformations are trivial, the ones that aren't I tested in situ if possible or ex situ if I lacked the configuration for it. For checking any booboos I defer to your capable eyes and the test-suite.

The shell on modern systems (sh) is a link to bash. So all of the testing I do uses bash. It'd be pretty easy for me to use some bash-isms then have the scripts break on a system with the Bourne shell.

As you say, there are several places that the scripts may use bash extensions. One example is getopts. So unless bash isn't available, I don't see an issue.

gonzoleeman (changes requested):
This is a lot of change just to switch shells. What is the problem with using bash, please? All of these changes mean it'll be easy for issue (bugs) to be added. I could handle that if we were fixing a bigger bug.

gonzoleeman:
I'm sorry I seem to have discouraged you to the point where you withdrew your pull request.

If you fell up to it some time, please explain to me your desire to remove dependence on bash. I assume you're building some place that does not have bash?

Perhaps we can continue this discussion as an issue?

PR #363, opened 2022-08-29
iscsiuio: Use pthread_t for INVALID_THREAD

kraj:
pthread_t is opaque, therefore avoid compiler errors on musl when
compiling since pthread_t is not a plain old data type, like glibc

Signed-off-by: Khem Raj <raj.khem@gmail.com>


gonzoleeman:
Thank you!

PR #364, opened 2022-09-05
Small bug fixes

gonzoleeman:
Small general clean, in preparation for the possibility of moving to use of 'meson' as our build system:

- move utils/sysdeps/ to sysdeps/
- keep "version" in one place -- the top-level Makefile
- fix/improve kernel-doc and its use, used in libopeniscsiusr to generate man pages
- fix a few kernel-doc bugs found in the libopeniscsiusr header files
- deliver all man pages uncompressed

Note: the only visible/functional change should be the last one, i.e. no more delivering some of the man pages compressed, since that's not needed and was inconsistent.


usr/version.h
```diff
@@ -5,8 +5,12 @@
  * iSCSI tools version.
  * This may not be the same value as the kernel versions because
  * some other maintainer could merge a patch without going through us
+ *
+ * Version string should be set by the build system, else we have problems
```
eli-schwartz, line 9: You could do this with a small sed version.h.in -> version.h (similar to what Meson might do). Also by adding autoconf to the top level (AC_CONFIG_SUBDIRS) but that is quite possibly more complexity than it's worth for a build system that would then be phased out.

gonzoleeman: Actually, it seems to be working fine.I have it soft-defined in the top-level Makefile, so it can be overridden when building, and the Makefile passes it to all compilations via a define.

And there is no top-level autoconf for open-iscsi. Currently, most of it uses 'make', and just the 'iscsiuio' subdirectory uses autoconf.

eli-schwartz: Right, like I said you'd have to add it and it would probably be a lot more bother than it's worth.

PR #367, opened 2022-09-13
Don't generate initiatorname when cross-building

tianyuanhao:
Let it be generated by the `iscsi-init` service.


gonzoleeman:
Oops. I merged this into open-iscsi:use-meson-v1, but use-meson-v2 is the correct (new) branch.

But then I cherry-picked this commit into the correct branch. So we should be good.

Thank you!

PR #368, opened 2022-09-20
Use realloc instead of reallocarray

tianyuanhao:
Fix build errors when not using glibc.

### DIFF
--- a/libopeniscsiusr/session.c
+++ b/libopeniscsiusr/session.c
@@ -291,7 +291,11 @@ int iscsi_sessions_get(struct iscsi_context *ctx,
 	}
 	/* reset session count and sessions array length to what we were able to read from sysfs */
 	*session_count = j;
-	*sessions = reallocarray(*sessions, *session_count, sizeof(struct iscsi_session *));
+	/* XXX: asserts that there is no integer overflow */
+	assert(!(sizeof(struct iscsi_session *) &&
+		 *session_count > UINT_MAX / sizeof(struct iscsi_session *)));
+	*sessions =
+	    realloc(*sessions, *session_count * sizeof(struct iscsi_session *));
 
 out:
 	free(sids);


gonzoleeman:
Please explain the problem encountered that this fixes? Is it that reallocarray() does not exist on platforms without glibc, but realloc() does?

tianyuanhao:
> Please explain the problem encountered that this fixes? Is it that reallocarray() does not exist on platforms without glibc, but realloc() does?

I'm adding the `open-iscsi` package to `Buildroot` and I found that `uClibc` does not implement `reallocarray`.

libopeniscsiusr/session.c
```diff
@@ -291,7 +291,7 @@ int iscsi_sessions_get(struct iscsi_context *ctx,
 	}
 	/* reset session count and sessions array length to what we were able to read from sysfs */
 	*session_count = j;
-	*sessions = reallocarray(*sessions, *session_count, sizeof(struct iscsi_session *));
+	*sessions = realloc(*sessions, *session_count * sizeof(struct iscsi_session *));
```
gonzoleeman, line 294: Please add a brief comment to the effect that we are using realloc() instead of reallocarray(), and why, and we assume the multiplication of "count" times "size" won't overflow -- something reallocarray() checks for us and realloc() does not.

gonzoleeman: After that, I'll be glad to merge this.

gonzoleeman:
Note: I am planning on tagging a new version "real soon now", and I'd like to have this fix in it, if possible. If you don't get around to adding the comment I can do it myself in a follow-up commit. But I'd prefer it came from you, since it's your patch.

gonzoleeman:
Thank you. I don't think you need to assert that the pointer size to the session structure is zero, but it shouldn't hurt. You also didn't add the comment I wanted, but I'll do that myself.

PR #377, opened 2022-10-11
iscsid/iscsiuio: fix OOM adjustment

gonzoleeman:
For both the iscsid and iscsiuio daemons, they try to modify nice value and OOM adjustment value, so they have priority and will not be killed by the OOM-killer.

But the code incorrectly set the value to "-17" for modern linux systems, when the maximum is actually "-1000". While making the changes, use "/proc/self/..." instead of "/proc/<PID>/...", so we don't have to use getpid() nor print out the pathname. Now we either write "-16" to the old interface, "-1000" to the new interface, or we print a warning.

Several "log_debug()" calls that should have been warnings are changed to "log_warning()" calls in iscsid.

### DIFF
--- a/iscsiuio/src/unix/main.c
+++ b/iscsiuio/src/unix/main.c
@@ -227,7 +227,7 @@ int oom_adjust(void)
 			LOG_DEBUG("Could not set /proc/self/oom_score_adj to -1000: %s",
 				strerror(errno));
 	} else if ((fd = open("/proc/self/oom_adj", O_WRONLY)) >= 0) {
-		if ((res = write(fd, "-16", 3)) < 0)
+		if ((res = write(fd, "-17", 3)) < 0)
 			LOG_DEBUG("Could not set /proc/self/oom_adj to -16: %s",
 				strerror(errno));
 	} else
--- a/iscsiuio/src/unix/main.c
+++ b/iscsiuio/src/unix/main.c
@@ -201,35 +201,43 @@ static void daemon_init()
 	close(fd);
 }
 
-#define ISCSI_OOM_PATH_LEN 48
-
+/*
+ * make a best effort at ajusting our nice
+ * score and our OOM score, but it's not considered
+ * fatal if either adjustment fails
+ *
+ * return 0 on success of OOM adjustment
+ */
 int oom_adjust(void)
 {
 	int fd;
-	char path[ISCSI_OOM_PATH_LEN];
-	struct stat statb;
+	int res = 0;
 
-	if (nice(-10) < 0)
+	errno = 0;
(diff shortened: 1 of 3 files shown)


ericwmackay:
Much simpler. LGTM

iscsiuio/src/unix/main.c
```diff
@@ -201,35 +201,43 @@ static void daemon_init()
...
+	/*
+	 * try the modern method of adjusting our OOM score,
+	 * then try the old one, if that fails
+	 */
+	if ((fd = open("/proc/self/oom_score_adj", O_WRONLY)) >= 0) {
+		if ((res = write(fd, "-1000", 5)) < 0)
+			LOG_DEBUG("Could not set /proc/self/oom_score_adj to -1000: %s",
+				strerror(errno));
+	} else if ((fd = open("/proc/self/oom_adj", O_WRONLY)) >= 0) {
+		if ((res = write(fd, "-16", 3)) < 0)
```
kalampakas, line 230: Disabling oom value is -17, perhaps the approach is to not disable oom for iscsid entirely ?

kalampakas (commented):
I just think it should be -17 for the case of oom_adj according to the documenation if you really want to disable oom.

iscsiuio/src/unix/main.c
gonzoleeman: Agreed. Brain fart. Changed to "-17" and pushed.

gonzoleeman:
If no objections, I'll merge this and close the other pull request (#376)

ericwmackay:
Actually, could you explain the choice of -16?

To echo what @kalampakas is saying, the man page for proc indicates that setting -1000 with the newer oom score scheme is roughly equivalent to -17 with the old scheme.

-16 is within the valid score range and -17 is a reserved value that disables OOM killing entirely.

-1000 isn't a reserved value but still has the effect of completely disabling OOM killing, so the behavior seems closer to -17.

(It still seems fine to use -16, but I'd like to fully understand the *why*.)

ericwmackay:
> Actually, could you explain the choice of -16?
> 
> To echo what @kalampakas is saying, the man page for proc indicates that setting -1000 with the newer oom score scheme is roughly equivalent to -17 with the old scheme.
> 
> -16 is within the valid score range and -17 is a reserved value that disables OOM killing entirely.
> 
> -1000 isn't a reserved value but still has the effect of completely disabling OOM killing, so the behavior seems closer to -17.
> 
> (It still seems fine to use -16, but I'd like to fully understand the _why_.)

Oops, NVM. Needed to refresh my web page. Carry on :)

Issue #381, opened 2022-12-04
Can we clear host scan process's PR_SET_IO_FLUSHER flag to fix failed to scan all devices reported by target


wenchao-hao:
Hi all, I meet an issue which we failed to scan all devices reported by target.

I have 16 targets with 256 LUNs per target, for each target, there are 2 path(we can simulate it to set node's nr_session to 2). It means we should have 16 * 256 * 2 = 8192 scsi devices after login, and each iSCSI session should have 256 scsi devices. However, the actually added scsi device is less than that number.

The root reason is we failed to alloc scsi_device's durning scan host. In kernel path:

```
scsi_probe_and_add_lun
    scsi_alloc_sdev
        blk_mq_init_queue ==> returned -ENOMEM
```
After probing, I found the failed reason for blk_mq_init_queue is pcpu_alloc called by subfunctions  in blk_mq_init_queue returned NULL. When allocating percpu variables, we usually called with GFP_KERNEL, but the flag would be modified by current_gfp_context(the ___GFP_IO and ___GFP_FS bit is cleared because iscsid is marked as PF_MEMALLOC_NOIO). Which would make this allocation atomic, so it returned NULL on no space left.

Since the PF_MEMALLOC_NOIO is set by  https://github.com/open-iscsi/open-iscsi/pull/270. Can we clear the PR_SET_IO_FLUSHER flags for host scan process?

wenchao-hao:
@mikechristie @gonzoleeman

mikechristie:
It would be ok to clear it for the scanning thread, because the thread is not in that main IO code path that can't block on itself without causing deadlocks.

wenchao-hao:
> It would be ok to clear it for the scanning thread, because the thread is not in that main IO/login code path that can't block on itself without causing deadlocks.

Thank you for your reply, I just submitted a PR to address this issue, would you help to review it?

wenchao-hao:
fixed,would close

PR #382, opened 2022-12-05
iscsid: clear scanning thread's PR_SET_IO_FLUSHER flag

wenchao-hao:
commit 72949ef (iscsid: set PR_SET_IO_FLUSHER) set the iscsid's PR_SET_IO_FLUSHER flag to avoid deadlock. While we do not need to set this flag when scanning host.

If this flag is set for scanning thread, we may lost devices reported by target because of memory allocation failure.

Reference: #381 

### DIFF
--- a/usr/iscsi_sysfs.c
+++ b/usr/iscsi_sysfs.c
@@ -40,6 +40,7 @@
 #include "host.h"
 #include "iscsi_err.h"
 #include "flashnode.h"
+#include "iscsi_util.h"
 
 /*
  * TODO: remove the _DIR defines and search for subsys dirs like
@@ -1993,6 +1994,12 @@ pid_t iscsi_sysfs_scan_host(int hostno, int async, int autoscan)
 	} else if (pid == 0) {
 		/* child */
 		log_debug(4, "scanning host%d", hostno);
+
+		/*
+		 * The return value of init_thread_io_flusher would not
+		 * affect the scan flow, so just ignore it.
+		 */
+		set_thread_io_flusher(0);
 
 		snprintf(id, sizeof(id), ISCSI_HOST_ID, hostno);
 		sysfs_set_param(id, SCSI_HOST_SUBSYS, "scan", write_buf,
--- a/usr/iscsi_util.c
+++ b/usr/iscsi_util.c
@@ -30,6 +30,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/resource.h>
+#include <sys/prctl.h>
 
 #include "sysdeps.h"
 #include "log.h"
@@ -37,6 +38,10 @@
 #include "iface.h"
 #include "session_info.h"
 #include "iscsi_util.h"
+
+#ifndef PR_SET_IO_FLUSHER
+#define PR_SET_IO_FLUSHER 57
+#endif
 
 int setup_abstract_addr(struct sockaddr_un *addr, char *unix_sock_name)
 {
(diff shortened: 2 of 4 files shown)


wenchao-hao:
@mikechristie @gonzoleeman

usr/iscsi_sysfs.c
```diff
@@ -1994,6 +1999,15 @@ pid_t iscsi_sysfs_scan_host(int hostno, int async, int autoscan)
 		/* child */
 		log_debug(4, "scanning host%d", hostno);
 
+		if (prctl(PR_SET_IO_FLUSHER, 0, 0, 0, 0) == -1) {
+			if (errno == EINVAL) {
+				log_info("prctl could not clear scanning thread's PR_SET_IO_FLUSHER flag, because the feature is not supported in this kernel.");
+			} else {
+				log_error("prctl could not clear scanning thread's PR_SET_IO_FLUSHER flag due to error %s, we may lost devices reported by target if memory is low",
+					  strerror(errno));
```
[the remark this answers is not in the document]
wenchao-hao: > I think you want to move the prctl code in iscsid.c to iscsi_util.c then call it here and iscsid.

updated.

usr/iscsi_util.c
```diff
@@ -392,3 +397,26 @@ int iscsi_match_target(void *data, struct session_info *info)
...
+	if (prctl(PR_SET_IO_FLUSHER, val, 0, 0, 0) == 0)
+		return 0;
+
+	/*
+	 * prctl would return EINVAL if the kernel do not support PR_SET_IO_FLUSHER
+	 * so donot print error log if errorno is EINVAL to avoid unnecessary errorlog
+	 */
+	if (errno != EINVAL)
+		log_error("prctl could not %s thread's PR_SET_IO_FLUSHER flag due to error %s\n",
+			val ? "set" : "clear", strerror(errno));
```
mikechristie, line 420: One small nit. I noticed in your patches you sometimes use formatting like above where the second line doesn't match up with the "(" above it. We generally line it up. For example if you look at the code you replaced we in iscsid.c we line it up, so you normally want to follow that.

usr/iscsi_util.c
```diff
@@ -392,3 +397,26 @@ int iscsi_match_target(void *data, struct session_info *info)
...
+
+/*
+ * init thread's PR_SET_IO_FLUSHER flag
+ *
+ * val: 1 to set to set thread's PR_SET_IO_FLUSHER flag
+ *      0 to clear thread's PR_SET_IO_FLUSHER flag
+ *
+ * return: return 0 on success, else error number is returned
+ */
+int init_thread_io_flusher(int val)
```
mikechristie, line 409: Maybe set_thread_io_flusher is a better name, because we are actually overriding the default that was set when the process was created and it was initialized.

usr/iscsi_util.c
wenchao-hao: Yes, I was used to keep the width of a line in 100 characters. I would pay attention to avoid this when submit patches to open-iscsi.

usr/iscsi_util.c
wenchao-hao: updated

gonzoleeman: I believe he made this change, so resolving this conversation.

usr/iscsi_util.c
gonzoleeman: Since this was a "nit" (which he understands), I will resolve this one.

usr/iscsi_sysfs.c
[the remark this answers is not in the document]
gonzoleeman: I believe he made this change, as requested, so marking this as resolved.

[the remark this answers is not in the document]
gonzoleeman: I believe he made this change, as requested, so marking this as resolved.

gonzoleeman:
I believe this is ready to merge. @mikechristie, any remaining issues? (I didn't see any.)

mikechristie:
Looks ok to me.

wenchao-hao:
@gonzoleeman Shall merge this PR now?

PR #384, opened 2022-12-13
Build(meson): ensure binaries use our library.

gonzoleeman:
Ensure that iscsid, iscsiadm, and iscsistart link
against libopeniscsiusr.so. Also, rename meson
list of "usr/" source files to be less confusing.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -167,6 +167,7 @@ foreach a,v: test_arr
   e = executable('test_' + a,
     sources: v,
     include_directories: [lib_inc],
+    link_with: libiscsi_usr,
     dependencies: libiscsi_usr_dep,
     c_args: genl_cargs)
   test('test ' + a, e)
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -167,6 +167,7 @@ foreach a,v: test_arr
   e = executable('test_' + a,
     sources: v,
     include_directories: [lib_inc],
+    link_with: libiscsi_usr,
     dependencies: libiscsi_usr_dep,
```
eli-schwartz, line 171: libiscsi_usr_dep is a container object that specifies `link_with: libiscsi_usr`, so this is redundant and the executables already link to libiscsi_usr.

meson.build
```diff
@@ -190,6 +191,7 @@ foreach k,v: iscsi_usr_arr
     include_directories: [usr_inc, main_inc, lib_inc],
     dependencies: usr_deps,
     c_args: genl_cargs,
+    link_with: libiscsi_usr,
```
eli-schwartz, line 194: Same here, because usr_deps includes libiscsi_usr_dep.

eli-schwartz (changes requested):

gonzoleeman:
Agreed, and thank you for looking at this. There is no need for this request, so I'm closing it.

Issue #387, opened 2023-01-03
iscsid failed to recover session if another error happened when iscsid is reopening the session.


wenchao-hao:
Here is the log:

```
Jan 03 03:57:25 fedora iscsid[40456]: iscsid: Kernel reported iSCSI connection 3:0 error (1020 - ISCSI_ERR_TCP_CONN_CLOSE: TCP connection closed) state (3)
Jan 03 03:57:25 fedora kernel:  connection3:0: detected conn error (1020)
Jan 03 03:57:28 fedora kernel:  connection3:0: detected conn error (1020)
Jan 03 03:57:32 fedora iscsid[40456]: iscsid: Received iferror -107: Transport endpoint is not connected.
Jan 03 03:57:32 fedora iscsid[40456]: iscsid: can't set operational parameter 15 for connection 3:0, retcode -107 (0)
Jan 03 03:57:34 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:36 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:38 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:40 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:42 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:44 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:46 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
Jan 03 03:57:48 fedora iscsid[40456]: iscsid: can't bind conn 3:0 to session 3, retcode -107 (115)
```

iscsid would failed to recover this session if error happened when it is reopening an session.

wenchao-hao:
The analysis is like following:

```
time
 |
 |      iscsid                                 kernel                          target
 |  
 |                                                                         close socket
 |                                                                   <-----------------------
 |                                     iscsi_conn_error_event()
 |                                     with error type
 |                                     ISCSI_ERR_TCP_CONN_CLOSE
 |                                    <------------------------
 |  
 |  session_conn_reopen()
[...]

PR #388, opened 2023-01-03
iscsid: stop connection for recovery if error is not timeout in iscsi_login_eh

wenchao-hao:
When iscsid is reopening a connection, and the reopen process has succeed to call bind_conn and comes to iscsi_session_set_params() to set parameters. If the iscsi target trigger another error event(such as close the socket connection between initiator and target) at this time, kernel would perform the error handler and set connection's state to ISCSI_CONN_FAILED, and set kernel iscsi_cls_conn->flags' ISCSI_CLS_CONN_BIT_CLEANUP bit. Which would make iscsid's iscsi_session_set_params() failed with ENOTCONN, so iscsi_login_eh() would be called by iscsid to handle this error.

Now iscsid see conn->state is ISCSI_CONN_STATE_XPT_WAIT and session->r_stage is R_STAGE_SESSION_REOPEN, so it would call session_conn_reopen() with do_stop set to 0, which would not trigger kernel to call iscsi_if_stop_conn() to clear kernel data struct iscsi_cls_conn->flags' ISCSI_CLS_CONN_BIT_CLEANUP bit.

The reopen would fall into an infinite cycle which looks like following:

iscsi_conn_connect -> bind_conn(failed with ENOTCONN)

         ^                     |
	 |                     |
	 |                     v

    session_conn_reopwn(with do_stop set to 0)

The phenomenon is iscsid would always report log "can't bind conn x:0 to session x, retcode -107 (115)" and the session would not recovery.

Fix this issue by checking error type in iscsi_login_eh(), if the error type is not timeout, make sure we would call session_conn_reopen() with do_stop set to STOP_CONN_RECOVER.

Fix https://github.com/open-iscsi/open-iscsi/issues/387


usr/initiator.c
```diff
@@ -735,8 +735,13 @@ static void iscsi_login_eh(struct iscsi_conn *conn, struct queue_task *qtask,
...
 			}
-			/* timeout during reopen connect. try again */
-			session_conn_reopen(conn, qtask, 0);
+			/*
+			 * stop connection for recovery if error is not
+			 * timeout
+			 */
+			if (err != ISCSI_ERR_PDU_TIMEOUT && err != ISCSI_ERR_TRANS_TIMEOUT)
+				stop_flag = STOP_CONN_RECOVER;
+			session_conn_reopen(conn, qtask, stop_flag);
```
mikechristie, line 744: Nice debugging. We will not get a ISCSI_ERR_PDU_TIMEOUT error in this code path will we?

wenchao-hao: > Nice debugging. We will not get a ISCSI_ERR_PDU_TIMEOUT error in this code path will we?

Yes, ISCSI_ERR_PDU_TIMEOUT would not happened if conn->state is ISCSI_CONN_STATE_XPT_WAIT.
My origin purpose is to handle all timeout sense. But it seems unnecessary, I would update

wenchao-hao: > Nice debugging. We will not get a ISCSI_ERR_PDU_TIMEOUT error in this code path will we?

@mikechristie updated

gonzoleeman:
It looks like this is ready, if Mike is satisfied. Your change looks like what he asked for.

mikechristie:
Patch looks ok to me.

PR #394, opened 2023-02-04
Avoid forcing remote-fs-pre after network-online if possible.

gordonmessmer:
Currently, when iscsi.service is installed, it creates an ordering dependency that forces network-online to start before remote-fs-pre.target, which delays boot even when /var/lib/iscsi/nodes is empty and the iscsi.service won't be started.  This change moves the logic that determines whether iscsi will be started to an external service file so that the boot order dependency exists only when iscsi.service will be started.

### DIFF
--- a/etc/Makefile
+++ b/etc/Makefile
@@ -27,6 +27,7 @@ SYSTEMD_SOURCES		= $(addprefix systemd/,$(SYSTEMD_SOURCE_FILES))
 SYSTEMD_TEMPLATE_FILES	= iscsi-init.service.template \
 			  iscsid.service.template \
 			  iscsi.service.template \
+			  iscsi-starter.service.template \
 			  iscsiuio.service.template
 SYSTEMD_TEMPLATES	= $(addprefix systemd/,$(SYSTEMD_TEMPLATE_FILES))
 SYSTEMD_RULES_FILES	= ibft-rule-generator
--- a/etc/systemd/iscsi-starter.service.template
+++ b/etc/systemd/iscsi-starter.service.template
@@ -1,0 +1,13 @@
+[Unit]
+DefaultDependencies=no
+Before=sysinit.target iscsi.service
+RequiresMountsFor=/var/lib/iscsi/nodes
+ConditionDirectoryNotEmpty=/var/lib/iscsi/nodes
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=@SBINDIR@/systemctl start --no-block --job-mode=fail iscsi.service
+
+[Install]
+WantedBy=sysinit.target
--- a/etc/systemd/iscsi.service.template
+++ b/etc/systemd/iscsi.service.template
@@ -15,4 +15,4 @@ SuccessExitStatus=21 15
 RemainAfterExit=true
 
 [Install]
-WantedBy=remote-fs.target
+Also=iscsi-starter.service


gordonmessmer:
This change is being merged into Fedora: https://src.fedoraproject.org/rpms/iscsi-initiator-utils/pull-request/10

The discussion that prompted this change is here: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/47J2M6A3IVDXFFJF5QOGYLXTNQMW52JL/#47J2M6A3IVDXFFJF5QOGYLXTNQMW52JL

gonzoleeman:
Also, this commit ignores the meson build system. Using Makefiles and make is deprecated, though there's no harm in updating the Makefile as far as I can see.

You have to ensure that the DBROOT gets expanded that you add to your service file.

gordonmessmer:
Both meson and make builds appears to work as expected, now.

gonzoleeman:
It looks good now. I will try to pull your branch and test it today. Thank you!

gonzoleeman:
Now that I look at this in more details, I'm unsure about your pull request.

It is not normal to call "systemctl start" from a unit to start another unit.

And I do not see anyplace where you are relying on the network being up. open-iscsi's iscsi.service relies on the network being up, else how can it log into targets? In particular, I worry about this arrangement when the root disc is on open-iscsi.

gordonmessmer:
> It is not normal to call "systemctl start" from a unit to start another unit.

Yes, that's a fair point.  We discussed making this a generator initially, but generators run before "optional" filesystems like /var are mounted, and Fedora stores iscsi configs in /var.  So, there may be some systems where that arrangement wouldn't work properly.

However, I've put together a VM that mounts its root filesystem via iscsi, installed a version of the iscsi tools that provides this systemd service, and instructed dracut to include the service in the initrd.  That VM does continue to boot normally.  systemd-analyze critical-chain from before and after are attached.

[fedora-critical-chain-starter.txt](https://github.com/open-iscsi/open-iscsi/files/10670998/fedora-critical-chain-starter.txt)
[fedora-critical-chain-stock.txt](https://github.com/open-iscsi/open-iscsi/files/10670999/fedora-critical-chain-stock.txt)

Perhaps an alternate approach would be to use a generator, and to instruct dracut to include iscsi support in the initrd if either the root fs is iscsi, or if any component of /var/lib/iscsi is on a different filesystem than root...

What do you think?

gordonmessmer:
... after further testing: including iscsi-starter.service in dracut doesn't seem to be necessary.  As far as I can tell, the iscsi.service in the initrd is simply enabled, even if it isn't in the host system.  So, everything works fine for root-fs-on-iscsi with Fedora's change, and without modifying dracut.

gonzoleeman:
I apologize for not replying to this, but I've been busy on another project (for open-iscsi), and I have also been thinking about this a bit.

I think I have to test it myself to see how our distribution handles this new service. Note that you don't just add a new service, but you modify the existing iscsi.service as well.

gordonmessmer:
No problem.  I also had a busy quarter at work, and haven't been able to come back to this for a bit.

If you'd like, I can implement an equivalent change as a generator instead and open a separate PR for that, for comparison.  The only drawback is that while it's a "cleaner" approach, it wouldn't work when configuration files are in /var (as they are in Fedora and RHEL), so if we go that way we either allow those distributions to diverge further from upstream, or we have to encourage them to use /etc instead of /var.

PR #396, opened 2023-02-13
SHA3 is not supported by libressl

ffontaine:
Disable SHA3 with libressl as it is not supported resulting in the following build failure:

```
/home/buildroot/autobuild/instance-3/output-1/host/lib/gcc/or1k-buildroot-linux-gnu/11.3.0/../../../../or1k-buildroot-linux-gnu/bin/ld: iscsid.p/usr_auth.c.o: in function `auth_hash_init': auth.c:(.text+0x7bc): undefined reference to `EVP_sha3_256'
```

Fixes:
 - http://autobuild.buildroot.org/results/48a4bddc355956733d712214797350cca8e111d9

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


gonzoleeman (changes requested):
I have no objection to making open-iscsi work with non-standard libraries, but I don't like having to name a define from one of those libraries to know if we should support SHA3 or not.

I am no encryption expert at all, but please explain to me why this library doesn't support SHA3-256? Looking at the sources of the latest version, it appears it supports some versions of SHA3. Also, when and why would this alternate library be used, instead of libopenssl?

Seems like both libopenssl and this one should have a way of telling clients which ciphers they support.

At the very least, if we can't make this test cleaner, we could at least hide that ugliness in an include file, e.g.:

/* libressl, an alternate openssl library, does not support SHA3-256, so test for that */
#ifdef LIBRESSL_VERSION_NUMBER
#define SHA3_256_NOT_SUPPORTED
#endif

then of course use that define for your checks.

ffontaine:
[libressl](https://www.libressl.org) is a fork of openssl. It is mostly used (and maintained) by OpenBSD. buildroot and OpenWRT also support it. To my knowledge, SHA3 is not supported because it was added to openssl after the fork: https://github.com/libressl/portable/issues/455. 

I'll update the PR to use meson [has_function](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_function).

gonzoleeman:
Merged. Thank you!

PR #398, opened 2023-03-01
Make some things optional

gonzoleeman:
These changes make building for iSNS or SLP discovery optional, using the `isns_suported` and `slp_supported` meson options. The defaults for both of these is "true". This means that, by default, there is no functional change.

The first two commits are cleanup, and do not change the build at all. The 3rd commit is where the actual build changes are contained. The last 2 commits update documentation.

Note that the Makefile infrastructure is not updated for these new options, since that build system is deprecated.

Note also that I may change the default to SLP support to "false" at some time in the future, since it is only partially implemented, and therefore never used as far as I know. But that will be later.

### DIFF
--- a/README
+++ b/README
@@ -81,7 +81,7 @@ As of today, the Open-iSCSI Initiator requires a host running the
 Linux operating system with kernel.
 
 The userspace components iscsid, iscsiadm and iscsistart require the
-open-isns library.
+open-isns library, if iSNS support has been chosen at build ttime.
 
 If this package is not available for your distribution, you can download
 and install it yourself.  To install the open-isns headers and library
@@ -205,8 +205,8 @@ options discussed earlier).
 Configuration is contained in directories for:
 
 	- nodes
-	- slp
-	- isns
+	- slp (if supported)
+	- isns (if supported)
 	- static
 	- fw
 	- send_targets
@@ -1254,7 +1254,8 @@ or discovered through the discover daemon iscsid.conf params
 (discussed in 7.1.2).
 
 If your distro does not have systemd units for iSCSI, then you will have
-to start the daemon and log into the targets manually.
+to start the daemon and log into the targets manually. You can also
+compile open-iscsi without systemd support, if you wish.
 
 
 7.1.1 iSCSI startup using the init script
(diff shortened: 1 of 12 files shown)


mikechristie:
I think you can just delete the SLP code. It's been on the README's TODO since the project was created so I think it's safe to say it's not happening :)

It was one of those things where linux-iscsi supported it. We planned to do it so users could transition. No one ever asked for it though, so it never got done. I think someone might have made a tool that does SLP calls and would create node db records using iscsiadm, but because no user requested SLP that never got merged as well.

gonzoleeman:
Withdrawing this pull request, for now. I will submit separate requests to (1) remove SLP support, and (2) make iSNS support optional.

gonzoleeman:
Actually closing.

mikechristie:
You have to leave the kernel code. iscsi_if.h falls under kernel code so you wouldn't change it. If some other tools, like qlogic's management tool (completely different userspace code compared to open-iscsi/usr and uses bsg and iscsi_if stuff directly), did support SLP iscsi discovery when talking directly to the kernel then it still works for them. This doesn't really affect open-iscsi/usr since we never added support for it.

You just drop the SLP code from the open-iscsi userspace tools. For example, iscsiadm/iscsid never supported ISCSI_TGT_DSCVR_SLP, so you don't need to do anything since there is no open-iscsi/usr code.


You would keep ISCSI_NET_PARAM_IPV4_DHCP_SLP_DA_EN in iface.c. It's not doing iscsi SLP discovery. It doesn't require outside libs since it just passes the info down to the card's fw which does all the work, so for what you are trying to accomplish you are ok still.

You would delete things like disocovery.c:slp_discovery and the slp code in iscsiadm.c.

You should be able to just delete the slp code from idbm.c as well. We never do anything with the slp fields in drec.slp, because iscsiadm -m discovery -t slp will return failure and never create slp records (/var/lib/iscsi/slp will always be empty).

If you just delete the idbm.c slp  code like:

idbm_recinfo_config(db->dinfo_slp, f);
idbm_recinfo_discovery(&db->drec_slp, db->dinfo_slp);
idbm_discovery_setup_defaults(&db->drec_slp, DISCOVERY_TYPE_SLP);
idbm_slp_defaults
...

that sets things up and requests them, idbm_recinfo_config will just ignore them in existing records. But, we should not have any slp records because iscsiadm never supported it so it should be an easy upgrade. Also we should never have slp for NODE_DISC_TYPE, so "slp" can just be removed there.

PR #401, opened 2023-03-08
Remove unused SLP Discovery code

gonzoleeman:
Support for SLP Discovery of targets was never fully implemented, and likely never will be, so remove stuff that supports it.

Note that iscsi_if.h was not functionally changed, since it must match the kernel. Also, the code for NFS SLP, a different beast, is still present.


gonzoleeman:
@mikechristie If you have time I'd really appreciate a review. I tested ST node discovery records work between the code before and after this change.

All other reviews are welcome, as well.

include/iscsi_if.h
```diff
@@ -104,6 +104,10 @@ enum iscsi_uevent_e {
 enum iscsi_tgt_dscvr {
 	ISCSI_TGT_DSCVR_SEND_TARGETS	= 1,
 	ISCSI_TGT_DSCVR_ISNS		= 2,
+	/*
+	 * SLP Discovery has never been implemented,
+	 * but this is kept for compatability with kernel code
+	 */
```
mikechristie, line 110: I would not change this, because it's supposed to mirror the kernel's iscsi_if.h. There is only minor compat changes to allow building it different userspace setups.

libopeniscsiusr/idbm.c
```diff
@@ -447,8 +447,6 @@ static int _idbm_iface_rec_link(struct iscsi_iface *iface,
 			 IDBM_SHOW, num, _CAN_MODIFY);
 		_rec_str(IFACE_DHCP_VID_STR, recs, iface, dhcp_vendor_id,
 			 IDBM_SHOW, num, _CAN_MODIFY);
-		_rec_str(IFACE_DHCP_SLP_DA, recs, iface, dhcp_slp_da, IDBM_SHOW,
-			 num, _CAN_MODIFY);
```
mikechristie, line 451: What was the reason for removing this one? Is the idea that it doesn't work or maybe not used like how the open-iscsi/usr might be?

include/iscsi_if.h
gonzoleeman: I will remove that comment, as a new commit. If you'd prefer I resubmit the branch, instead, let me know.

libopeniscsiusr/idbm.c
gonzoleeman: No, just an error on my part. I will fix this.

Since there are two fixes now to my submission, I will just fix them both and force-push the update.

gonzoleeman:
@mikechristie I pushed the updated branch.

mikechristie:
Looks ok to me.

PR #404, opened 2023-03-16
Make UID the default auth type for IPCs

ericwmackay:
Modern security practices include checking the originating UID of an IPC. But not the username, as there are many valid scenarios where a user database may not exist or the root account may not be explicitly named "root".

The mode to check UID and username is still functional, but is now on the path to deprecation. Related function names and enums now include the term "legacy" to indicate this deprecated status.

The 'iscsid.ipc_auth_uid' setting in iscsid.config now defaults to 'Yes', rather than introducing a new key. Users sometimes keep old config files when upgrading packages, and this ensures behavior remains the same if this setting was explicitly defined in an old config.

### DIFF
--- a/etc/iscsid.conf
+++ b/etc/iscsid.conf
@@ -31,10 +31,11 @@
 # iscsid.safe_logout = Yes
 
 # Only require UID auth for MGMT IPCs, and not username.
-# Useful if you want to run iscsid in a constrained environment.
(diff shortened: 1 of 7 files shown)


gonzoleeman:
Thank you!  BTW I did a minor edit on the commit message, removing a couple of blank lines.

Issue #408, opened 2023-04-14
idbm write error on update when records with/without tpgt tag clash


cleech:
This is an old issue that for some reason the patches didn't get merged upstream.
I'm fixing that, so I'm just opening this to document the issue and reference it from a pull request that I'm about to open.

There's a particular series of commands that the storage manager from the Red Hat virtualization product was issuing, that would result in an error trying to update a node record.

It also depends on the return order of readdir, which is undefined and file-system dependent.  I think it only ever reliably happened with XFS.
You can force a predictable readdir ordering that is likely to reproduce this by using the spd_readdir LD_PRELOAD from e2fsprogs. 
[https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/plain/contrib/spd_readdir.c]

This script reproduces the sequence of discovery, new static record without tgpt, attempt to update record.   When it fails, the update will report a database error.
```
== Setting 172.18.210.9 node.startup -> manual
iscsiadm: iface iter could not stat /var/lib/iscsi/nodes/iqn.1994-05.com.redhat:rhel-storage-37/172.18.210.9,3260.
iscsiadm: Could not execute operation on all records: encountered iSCSI database failure
```

```
#!/bin/sh
#
# The spd_readdir LD_PRELOAD can help reproduce this on filesystems other than XFS
# https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/plain/contrib/spd_readdir.c
# gcc -shared -ldl -fPIC spd_readdir.c -o spd_readdir.o

export LD_PRELOAD=${PWD}/spd_readdir.o

ISCSIADM=iscsiadm

TARGET=iqn.1992-01.com.example:target.0 
PORTAL=192.168.122.1:3260

delete_all () {
	p=$1
	echo "== Clearing DB of all node records"
	${ISCSIADM} -m node -o delete

	echo "== Clearing DB of all discovery records"
	${ISCSIADM} -m discoverydb -o show > /dev/null
	if [ $? -eq 0 ]; then
		${ISCSIADM} -m discoverydb -t st -p $p -o delete
	else
		echo "scsiadm: No discovery records found"
	fi
}

discovery () {
	portal=$1
	echo "== Performing SendTargets Discovery against ${portal}"
	${ISCSIADM} -m discoverydb -t st -p $portal -o new
	${ISCSIADM} -m discoverydb -t st -p $portal --discover
}

new_static () {
	target=$1
	portal=$2
[...]

PR #409, opened 2023-04-14
When writing DB record without a tpgt, take tpgt from existing record

cleech:
The problem is described in issue #408 

These patches split up idbm_write_rec to split out the logic from the old records without a tpgt in the path, and the new records that we should be using.  Then it updates the old record format writing to check for a clashing new style record first, and update that instead if it exists.


usr/idbm.c
```diff
@@ -2220,24 +2184,103 @@ static int idbm_rec_write(node_rec_t *rec, bool disable_lock)
...
+	idbm_print(IDBM_PRINT_TYPE_NODE, rec, 1, f);
+	fclose(f);
+free_portal:
+	free(portal);
+	return rc;
+}
+
+/*
+ * When the disable_lock param is true, the idbm_lock/idbm_unlock needs
+ * to be holt by the caller, this will avoid overwriting each other in
```
gonzoleeman, line 2278: typo: "holt" -> "held"

gonzoleeman (commented):
Chris: could you explain at a high level what this is actually about? Was there a bug or problem that this is fixing? I'm not even sure I understand "old" vs "new" records. Did open-iscsi used to not specify the target portal group?

As far as the changes, only one small typo in a comment. :) Your choice if you want to fix it.

usr/idbm.c
cleech: Hey, i just moved the typo :-)
I'll fix it

cleech:
> Chris: could you explain at a high level what this is actually about? Was there a bug or problem that this is fixing? I'm not even sure I understand "old" vs "new" records. Did open-iscsi used to not specify the target portal group?

Did you look at issue #408 ?  I don't know why I though an extended description deserved to be in an issue instead of the pull request, but I put the info over there.

iscsiadm will let you create a static node record with no tgpt, which gets saved under a slightly different path format.  Combining that with discovery can result in two records. Then, depending on the filesystem-dependant ordering of readdir, you can issue an update command that fails with a database error.

PR #410, opened 2023-04-14
iscsi-iname: make default IQN prefix configurable

cleech:
Packagers might want to change the default IQN prefix to use their own domain name information.  Provide a build time option for that.


cleech:
Thanks, Red Hat has always patched the iscsi-iname sources when packaging to use iqn.1994-05.com.redhat.  I really appreciate how much you've cleaned up the build system, and I'm working on cleaning up our packaging to drop these minor patches.

PR #411, opened 2023-04-14
sync iscsi_if.h with kernel

cleech:
Adds definition for ISCSI_UEVENT_DESTROY_SESSION_ASYNC

Replaces zero-length array with C99 flexible array member

### DIFF
--- a/include/iscsi_if.h
+++ b/include/iscsi_if.h
@@ -22,12 +22,11 @@
 #define ISCSI_IF_H
 
 #ifdef __KERNEL__
+#include <scsi/iscsi_proto.h>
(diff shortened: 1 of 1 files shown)


gonzoleeman:
This looks like mostly cleanup, plus converting ending arrays[0] with array[]. So does the kernel include file also need to change, to match?

cleech:
These changes all came from the kernel version, I'm just trying to get them closer to the same.

gonzoleeman:
Sometimes, I wonder why we have a version number between the kernel and iscsid. We never update it. :-/

PR #415, opened 2023-04-20
Remove incorrect "BindTo" line in iscsiuio.service

gonzoleeman:
The correct syntax is "BindsTo", but we do not
actually want to use this, because it would imply
that iscsid.service needs to be started _before_
iscsiuio.service, but that is backwards.

Found by: Martin Wilck @ SUSE


gonzoleeman:
Note: It looks like "BindTo=" is the old version, and "BindsTo=" is the new one. And systemd seems to still support the older "BindTo=", treating it just like "BindsTo=". But, again, we should not be Binding iscsiuio to to iscsid, since that means the wrong order of startup.

PR #425, opened 2023-09-22
discoveryd: fix additional format overflow warnings

ginggs:
Similar to #413, these were seen compiling with -O3


gonzoleeman:
Thank you.

PR #426, opened 2023-09-30
usr/idbm.c: fix musl build

ffontaine:
Fix the following musl build failure raised since version 2.1.9 and https://github.com/open-iscsi/open-iscsi/commit/7b571d76d6937a78c141630fc38c3c57c532466c:

```
../usr/idbm.c: In function 'idbm_rec_write_old':
../usr/idbm.c:2230:27: error: 'GLOB_ONLYDIR' undeclared (first use in this function)
 2230 |         rc = glob(portal, GLOB_ONLYDIR, NULL, &globbuf);
      |                           ^~~~~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/14fc1c139f055b5b1eaa6e04e327863c06176a7b

### DIFF
--- a/usr/idbm.c
+++ b/usr/idbm.c
@@ -45,6 +45,11 @@
 #include "fw_context.h"
 #include "iscsi_err.h"
 
+// GLOB_ONLYDIR is not defined under musl
+#ifndef GLOB_ONLYDIR
+#define GLOB_ONLYDIR	0x100
+#endif
+
 #define IDBM_HIDE	0    /* Hide parameter when print. */
 #define IDBM_SHOW	1    /* Show parameter when print. */
 #define IDBM_MASKED	2    /* Show "stars" instead of real value when print */


gonzoleeman (commented):
Please add a comment about why GLOB_ONLYDIR is being defined here, since it normally doesn't have to be defined (i.e. using libc), and I'll be glad to merge it.

gonzoleeman (changes requested):
Please add comment. (I forgot to set "request changes" on my previous review)

ffontaine:
Thanks for your review, I'll add the comment.
However, I got the following remark from one of buildroot maintainer:

> What is the impact on the application if the glob also matches files, not just directories? Perhaps it breaks the application behavior. Are you sure we can just blindly do this workaround?

Can you confirm that the application will not break if glob also matches files?

gonzoleeman:
> Thanks for your review, I'll add the comment. However, I got the following remark from one of buildroot maintainer:
> 
> > What is the impact on the application if the glob also matches files, not just directories? Perhaps it breaks the application behavior. Are you sure we can just blindly do this workaround?
> 
> Can you confirm that the application will not break if glob also matches files?

Looking at the code, this section is for handling old-style open-iscsi database (DB) records for nodes. The old-style records didn't have a target-portal-group, I believe (based on looking at the code, and discussions with Mike Christie a while back).

So can I say that nothing will ever break if files are found rather than directories? I believe so. The record names are much different than the directory names, and this section of code is looking for a particular name. So I believe it would only screw up if somebody created a file name that was the same as the directory being searched for, and that seems impossible.

gonzoleeman (approved):

PR #427, opened 2023-10-04
Iscsiuio fixes v2

gonzoleeman:
Multiple changes to iscsiuio to fix some instability, particularly when shutting down. This does not fix everything, but it's pretty close, I believe.

### DIFF
--- a/iscsiuio/src/apps/dhcpc/dhcpc.c
+++ b/iscsiuio/src/apps/dhcpc/dhcpc.c
@@ -258,7 +258,7 @@ static PT_THREAD(handle_dhcp(struct uip_stack *ustack))
 	s = ustack->dhcpc;
 
 	if (s == NULL) {
-		LOG_WARN("Could not find dhcpc state");
+		ILOG_WARN("Could not find dhcpc state");
 		return PT_ENDED;
 	}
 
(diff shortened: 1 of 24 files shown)


cleech:
Thanks Lee, I'll give these a look

gonzoleeman:
Hi @cleech . Any chance you've had a chance to look at this? FYI, there are still some changing coming with respect to thread handling at shutdown that a co-worker has authored. With these changes and those, Marvell has tested and was able to do 70 iterations of shutdown/restart without issue. FTR, that's about 69 more than it used to do. We will continue to analyze what happened on iteration 70. I fear there may still be some timing issues that can show up rarely, depending on CPU speed, system load, phase of the moon, etc.

cleech:
Sorry, been busy looking at some kernel side stuff with these same drivers!  Started back in on a review earlier this morning.

cleech:
OK, this all looks pretty good to me.  Thanks for all the work on that.

My only comment was on setting SO_LINGER for the netlink socket, because I'm not sure that does anything.  But it also doesn't hurt, so I'm OK with merging all this as is.

gonzoleeman:
@cleech thank you for the review! I added a comment about SO_LINGER possibly having no effect.

I plan on waiting another day or two, for possible feedback from Marvell, before merging.

njavali:
Lee, the changes look good and also verified on our test environment. 
Since /var/log/iscsiuio.log is removed, would all the iscsiuio logs be captured from stdout and would it not clutter the output. And syslog would be used for all background mode logging. Is that the correct understanding ?

gonzoleeman:
> Lee, the changes look good and also verified on our test environment. Since /var/log/iscsiuio.log is removed, would all the iscsiuio logs be captured from stdout and would it not clutter the output. And syslog would be used for all background mode logging. Is that the correct understanding ?

Yes, that is correct. The logging is currently set up to print to stdout or stderr (for error messages) if in foreground mode, and to go to the system log (with "iscsiuio" prepended to the messages), if in background mode.

The code before this change printed to the log file all the time (if logging was enabled, which it always is), and also to stdout if in debug mode. So that made debug mode very costly in terms of logging overhead.

The new method dumps to stdout when in foreground mode because that's how it's called by systemd, and that way journalctl can be used to look at the output. But yes, if a regular admin runs iscsiuio in the foreground, they will get all of the logging to stdout/stderr.

I believe this is how iscsid is currently working, as well.

PR #428, opened 2023-10-17
iscsiuio: manage threads better at shutdown

gonzoleeman:
This commit adds ability for iscsiuio to count
threads started, and to wait for them at shutdown, up to 10 seconds maximum. It also adds a few
messages during that shutdown time if threads are
slow to shut down.

Testing showed that 100 iterations of stop/start
of the iscsiuio service completed without error
with these changes.

### DIFF
--- a/iscsiuio/src/unix/main.c
+++ b/iscsiuio/src/unix/main.c
@@ -101,11 +101,42 @@ struct options opt = {
 };
 
 int event_loop_stop;
+/*
+ * The number of threads currently using event_loop_stop for synchronization purposes
+ * Each should lock/increment/unlock before starting is processing loop
+ * As each observes the stop flag being set it should lock/decrement/unlock
+ * This will allow the cleanup routine to not issue cancels to active threads
+ */
+static int event_loop_observers;
+static pthread_mutex_t event_loop_observers_mutex;
+
 extern nic_t *nic_list;
 
 struct utsname cur_utsname;
 
-/**
+/*
+ * event_loop_observer_add() - Increment the number of areas of code currently 
+ * observing event_loop_stop as an exit/shutdown mechanism
+ */
+void event_loop_observer_add(void)
+{
+	pthread_mutex_lock(&event_loop_observers_mutex);
+	event_loop_observers++;
+	pthread_mutex_unlock(&event_loop_observers_mutex);
+}
+
+/*
+ * event_loop_observer_add() - decrement the number of areas of code currently 
+ * observing event_loop_stop as an exit/shutdown mechanism
+ */
+void event_loop_observer_remove(void)
+{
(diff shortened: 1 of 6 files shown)


iscsiuio/src/unix/main.c
```diff
@@ -161,6 +193,28 @@ static void *signal_handle_thread(void *arg)
...
 	ILOG_INFO("terminating...");
 
+	/*
+	 * for debugging shutdown issues, let's wait 10 seconds, max,
+	 * to ensure all of our threads shutdown, since we have seen
+	 * issues where they may not do so.
+	 */
+	waitcount = WAITCOUNT_MAX;
+	while (waitcount--) {
+		sleep(1);
```
cleech, line 203: Would moving the sleep to the end of the loop prevent taking a 1 second sleep when none is needed?
Do you have an idea of what the average wait time here is in your testing?  Just curious.

cleech:
This looks like it should work, even if I'm not crazy about adding delays to shutdown in the magnitude of seconds.

iscsiuio/src/unix/main.c
gonzoleeman: I will look at moving the sleep(1). In answer to your question, we ran about 50 reboot cycles and never saw a delay of 2 seconds (with iSCSI root).

gonzoleeman: @cleech what do you think about this?

```
        ...
	waitcount = WAITCOUNT_MAX;
	while ((event_loop_observers > 0) && waitcount--) {
		sleep(1);
		if (event_loop_observers <= 0)
			break;	/* they finished while we were sleeping */
		ILOG_INFO("%d threads still polling event_loop_stop flag after %d seconds",
			  event_loop_observers, WAITCOUNT_MAX - waitcount);
	}
	if (event_loop_observers < 0)
		ILOG_DEBUG("Invalid observer count: %d", event_loop_observers);
	else if (event_loop_observers > 0)
		ILOG_ERR("%d unresponsive observers will be cancelled: %d",
			 event_loop_observers);
	cleanup();
	exit(EXIT_SUCCESS);
```
This has the advantage that if there are no observers at all when we first enter the loop, we will not sleep at all. But, in practice, I think it's likely the sleep(1) that gives the threads a chance to complete their runs and decrease the observer count. So the code checks again after waking up. I also changed the message about a negative observer count to a debug message, so we don't freak normal users out.

I also think we could likely change the maximum sleep to something like 5 seconds, but in practice (as I said before) we only once saw a sleep of 2 seconds.

If you approve of the above suggested change, I'll push an update.

gonzoleeman:
Crap. Now I realize I really need to update the version number of iscsiuio, since this was some major changes. :(

PR #432, opened 2023-12-04
Incorrect documentation for `iscsiadm -m session` print level

daniel-moulton:
The documentation states that the print level can be between 0 and 2, but then lists the levels as 1 to 3.
Is 

Excerpt provided below for context:
```
Print level can be 0 to 2.
			  1 = Print basic session info like node we are
			  connected to and whether we are connected.
			  2 = Print iSCSI params used.
			  3 = Print SCSI info like LUNs, device state.
```

I've updated the range of values to include 3, and provided a description for print level 0.

### DIFF
--- a/README
+++ b/README
@@ -486,7 +486,8 @@ Mode "session"
 			  but in session mode targetname and portal info
 			  is not passed in.
 
-			  Print level can be 0 to 2.
+			  Print level can be 0 to 3.
+			  0 = Print the running sessions.
 			  1 = Print basic session info like node we are
 			  connected to and whether we are connected.
 			  2 = Print iSCSI params used.


gonzoleeman:
Changes look good, but any chance you can add a line about level 3? Level 3 adds device and host info.

And yes, the man page should have this too. Perhaps subcommands (i.e. the "mode") should have their own man page, much like git subcommands have their own man pages. Hmmm.

daniel-moulton:
There's already a line for level 3: `3 = Print SCSI info like LUNs, device state.`

Is this description sufficient or does it need updating?

Think there's ups and downs of either approach, having mode-specific man pages does make sense though - would speed up finding the information you need but might not warrant the overhead of creating the individual pages.

gonzoleeman:
You are correct, 3 is already listed. Thanks!

Issue #433, opened 2023-12-06
Use of deprecated inet_aton, inet_ntoa fuctions


martinhoyer:
As I understand, inet_ntop and inet_pton are the successors for these functions, which have been flagged by [Fedora linter](https://github.com/rpminspect/rpminspect-data-fedora/blob/0195c42882161cde10a8ab4fb39ca795014b1d3a/fedora.yaml#L745C16-L745C16) 

The functions can be found in the following files.
inet_aton:
/usr/iscsi_net_util.c

inet_ntoa:
/iscsiuio/src/uip/uip_arp.c
/iscsiuio/src/unix/iscsid_ipc.c
/iscsiuio/src/unix/libs/cnic.c
/iscsiuio/src/unix/nic.c

gonzoleeman:
See pull request #435

PR #435, opened 2023-12-19
Stop using deprecated inet_aton and inet_ntoa

gonzoleeman:
The use of inet_aton is in usr/icsi_net_util.c, and the use of inet_ntoa is in iscsiuio/src/...
Use the modern versions, inet_pton and inet_ntop,
which support IPv6, are used instead, but they
require a buffer. See open-iscsi issue #433.


martinhoyer:
Thanks! Closes #433

PR #437, opened 2023-12-29
Using vendor defined directories for configuration files besides user/admin defined configuration files.

schubi2:
This PR should fulfill the requirements regarding "Configuration Files Specification" of the
Linux Userspace API Group:
https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md


gonzoleeman:
I have some issues with this, and will have to look it over more closely.  What is "sestatus.conf"? It is mentioned in docs but doesn't seem used?

Of course this adds a dependency on this new config library libeconf, but that seems reasonable. But why not use it, then, for parsing, and why make this and ifdef around using a "vendor directory"? Also, since I know nothing yet about libeconf, how does this merging actually work? Is the "vendor" config file only used if /etc/iscsi/iscsid.conf is not present? What if they are both present, i.e. which takes precendence? Are any configuration items merged, or is it one or the other? And is a vendor "directory" supported, i.e. a dot-d directory with config files under it (perhaps per-user, or ???).

I believe the README should at least explain some of this stuff, since we don't have an "iscsid.conf" man page.

gonzoleeman:
Note: I also do not see any installation of a vendor-dir config file, since the specification states that the configuration should work even if no configuration file is present in /etc. So shouldn't we be installing an iscsid.conf in /usr some place? And, speaking of that, there doesn't seem to be any default vendor directory. So what would you propose for the default location?

Lastly, it does not look like you modified the stale but still existing autoconf code to handle this option. Was that intentional?

gonzoleeman:
I pulled your branch and tested it out. I have not yet tested the functionality, i.e. I did not create more than the normal config file to see what happens. I was just testing the functionality around building the software.

As I said before, if one chooses this "feature", it should "just work" with default values. There is no default vendor dir, and there should be. Also, it looks like you defined the new functionality this way:

- a new "feature" called "econ" is available, but not sure what the default value is, since you didn't set that
- setting the new "econ" feature to "enabled" actually doesn't do anything, by default, unless "vendordir" is set

So problems are:

- vendordir needs a default value, so it shouldn't need to be set unless one wants to override that default
- "econf" needs a better name as a feature, like "multi-conf", or something at least slightly intuitive
- you need to install a config file in vendordir as part of the install, just like we do in /etc/iscsi
- the autoconf code needs to be updated, unless you are proposing to remove it (which will involve more discussion, but perhaps less work in the long run)

For the record, I think this is an ok idea, solving a problem that doesn't really exist. But I'm open to allowing it.

gonzoleeman:
Adding to previous post, also needed is something in the README a little better than what you proposed.

I just thought of one more possible functional surprise if we do this: right now, if a user comments out a value in /etc/iscsid.conf, they will get the default (which is listed as a comment in said file). But after this change, they will instead get whatever is set in <vendordir>/iscsid.conf, if set there, else the default. That might mean changing the format of iscsid.conf, or at least adding comments to this effect.

I really appreciate your submission, but it seems there might be more to this than what appeared at first glance, for me.

schubi2:
Hi Lee,
thank you for your really good suggestions. I see your points and I am working on it ...... :-)

schubi2:
> I have some issues with this, and will have to look it over more closely. What is "sestatus.conf"? It is mentioned in docs but doesn't seem used?

Yes, that has been a typo. Fixed....

> 
> Of course this adds a dependency on this new config library libeconf, but that seems reasonable. But why not use it, then, for parsing, and why make this and ifdef around using a "vendor directory"? Also, since I know nothing yet about libeconf, how does this merging actually work? Is the "vendor" config file only used if /etc/iscsi/iscsid.conf is not present? What if they are both present, i.e. which takes precendence? Are any configuration items merged, or is it one or the other? And is a vendor "directory" supported, i.e. a dot-d directory with config files under it (perhaps per-user, or ???).
>

libeconf will be used for parsing and merging vendor and user/admin defined iscsid.conf files. They way how it will be
done has been defined by the Linux Userspace API Group:
https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md
and has been implemented by libeconf.
 
> I believe the README should at least explain some of this stuff, since we don't have an "iscsid.conf" man page.

Good point.  I have added for information there.

schubi2:
> So problems are:
> 
> * vendordir needs a default value, so it shouldn't need to be set unless one wants to override that default

Hm, I think we should keep the current behaviour. If no vendordir has been defined iscsid.conf  in /etc/ we be used only.
If vendor dir has been defined, vendor and user/defined iscsid.conf  files will be parsed and evtl. merged.
I think we should NOT provide any default vendor directory because (as far I know) linux distributions are using different vendor directories. E.g. SUSE are using /usr/etc.

> * "econf" needs a better name as a feature, like "multi-conf", or something at least slightly intuitive

Yes, you are right, this flag is confusion and is not needed at all. So, libeconf is needed, if vendordir has been
defined. There should not be needed an extra econf option. I have removed it.

> * you need to install a config file in vendordir as part of the install, just like we do in /etc/iscsi
> * the autoconf code needs to be updated, unless you are proposing to remove it (which will involve more discussion, but perhaps less work in the long run)

I have fixed these two points. Thanks !

schubi2:
> Adding to previous post, also needed is something in the README a little better than what you proposed.
> 

Yes, that's a good point. I have added more information there.

> I just thought of one more possible functional surprise if we do this: right now, if a user comments out a value in /etc/iscsid.conf, they will get the default (which is listed as a comment in said file). But after this change, they will instead get whatever is set in /iscsid.conf, if set there, else the default. That might mean changing the format of iscsid.conf, or at least adding comments to this effect.
> 

Please have a look to 
https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md
The Linux User Group has had the same thoughts and have given the answers for that. Ähm, hopefully :-)

gonzoleeman:
FWIW I'm still looking at this.

Open-iscsi will soon have a need to exist in "transactional" OSes, where some directory trees are read-only, so both config files and database files may need to be able to migrate elsewhere.

schubi2:
Have you had time meanwhile ?

PR #441, opened 2024-02-23
Fix setpriority issue

gonzoleeman:
This fixes an issue in session_increase_wq_priority(). This function is called when setting up a session and host to increase the workqueue priority. The workqueue is named (by the kernel) "iscsi_q_HOSTNO", where HOSTNO is the host number. It previously did this by scanning /proc/PROCNO/stat files, but that was error prone and cumbersome/slow.  It used the "setpriority()" call on the process ID for the worker thread to do this, if it could find the process entry.

Instead, write to the "nice" value for the workqueue in sysfs, which is simpler, faster, and more reliable.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -22,7 +22,6 @@
 #include <sys/time.h>
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <sys/resource.h>
 #include <unistd.h>
 #include <string.h>
 #include <stdio.h>
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1389,85 +1389,64 @@ static void session_conn_recv_pdu(void *data)
 
 static void session_increase_wq_priority(struct iscsi_session *session)
 {
-	DIR *proc_dir;
-	struct dirent *proc_dent;
+	char nice_pathname[PATH_SIZE];
 	struct stat statb;
-	char stat_file[PATH_SIZE];
-	char sbuf[1024];	/* got this from ps */
-	int pid, stat_fd, num_read;
-	char *proc_name, *proc_name_end;
-	uint32_t host_no;
+	int nice_fd;
+	char nice_value_buf[10];	/* large enough for "-20" */
+	int buflen;
+
+	log_debug(5, "Trying to increase workqueue priority: sid %u host no %u",
+		  session->id, session->hostno);
 
 	/* drivers like bnx2i and qla4xxx do not have a write wq */
-	if (session->t->caps & CAP_DATA_PATH_OFFLOAD)
+	if (session->t->caps & CAP_DATA_PATH_OFFLOAD) {
+		log_debug(5, "Offload host detected: no workqueue");
 		return;
-
-	proc_dir = opendir(PROC_DIR);
-	if (!proc_dir)
-		goto fail;
-
-	while ((proc_dent = readdir(proc_dir))) {
-		if (!strcmp(proc_dent->d_name, ".") ||
-		    !strcmp(proc_dent->d_name, ".."))
-			continue;
-		if (sscanf(proc_dent->d_name, "%d", &pid) != 1)
-			continue;
-
-		memset(stat_file, 0, sizeof(stat_file));
-		sprintf(stat_file, PROC_DIR"/%d/stat", pid);
-		if (stat(stat_file, &statb))
-			continue;
-
-		if (!S_ISREG( statb.st_mode))
-			continue;
-
-		stat_fd = open(stat_file, O_RDONLY);
-		if (stat_fd == -1)
-			continue;
-
-		memset(sbuf, 0, sizeof(sbuf));
-		num_read = read(stat_fd, sbuf, sizeof(sbuf));
-		close(stat_fd);
-		if (num_read == -1)
-			continue;
-		if (num_read == sizeof(sbuf))
-			sbuf[num_read - 1] = '\0';
-		else
-			sbuf[num_read] = '\0';
-
-		/*
-		 * Finally match proc name to iscsi thread name.
-		 * In newer kernels the name is iscsi_wq_%HOST_NO.
-		 * In older kernels before 2.6.30, it was scsi_wq_%HOST_NO.
-		 *
-		 * We only support newer kernels.
-		 */
-		proc_name = strchr(sbuf, '(') + 1;
-		if (!proc_name)
-			continue;
-
-		proc_name_end = strchr(proc_name, ')');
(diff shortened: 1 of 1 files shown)


gonzoleeman:
I added a bit of debugging code -- perhaps too much? Comments welcome on that as well as on my approach, in general. Hannes suggested letting the kernel part of the initiator code set the workqueue priority when it received a request from iscsi via the netlink socket, but that seems like more work (and has the same result, I believe).

gonzoleeman:
@mikechristie any comments? I believe you wrote and update the code that sets the workqueue thread nice value. On modern kernels the search for "iscsi_q_%HOST_NO" is failing often, but using the sysfs workqueue "nice" attribute works and seems more reliable. But I believe the workqueues have changed a bit in 6.6 anyway.

mikechristie:
> Even the current function (that scans the proc table) has notes that it ignores older kernels.

When you add a feature you don't need to support older kernels. If you have an existing feature, you don't want to just remove it from users.

> I really don’t want to improve the code by having both new and old functionality present.

It's just a part of balancing your role  where you support users vs wanting clean code.

> And even if, on older kernels, it fails, that would be better than it failing on new ones. It just effects the nice 

Unless a user is using it and now hits a regression.

> value for the workqueue. And the old method actually hacked on the nice vale for the process, not the workqueue itself.

In older kernels, workqueues worked a lot differently so you had to set the nice value of the process. There was not sysfs file for example and the way we managed threads was a lot different. In the original wq code we had dedicated thread and it was less dynamic.

If you want to make the argument, it hasn't worked for so long that there are probably no users, then I can maybe see that.

Maybe we could go even further then. I think one of the reasons we needed it was not for perf, but to avoid some sort of priority inversion bug. However, this has been broken for a while. I haven't watched the lists and was not doing distro support stuff for many years now, but if no one has hit the bug and no one has mentioned perf regressions, maybe we don't even need the code anymore and can remove it?

Do what you feel is best because I only have the one user I work with primarily. You guys see more. Either we have users and don't want to add a regression for them, or we don't have users and can kill the old code and maybe even just remove the code all together.

mikechristie:
> If you want to make the argument, it hasn't worked for so long that there are probably no users, then I can maybe see that.

Oh  yeah, if we also say maybe this only worked for so few kernels that no one used it, then I can see that as well.

My concern is just if there are users and will they care if there is a regression. If they don't care or there is no regression I don't care. Do what makes ya happy :)

gonzoleeman:
I have done quite a bit of benchmarking using a set of fio tests. I did some on a 5.14 kernel, which has the sysfs interfaces, and some on 5.3 kernel which does not.

On the 5.3 kernel, it seems that my changes would mean that the workqueue thread did not get reniced. In testing there, I saw quite a bit of variability in the test using 16 parallel 1G random writes. The other two tests I ran were a single process doing random 4k writes, and single process random 1Mb writes. Both of those showed little changes in IOPs or BW.

So, bottom line, there seems to be some regression here. But this is an old kernel. How many folks get the latest open-iscsi and put it on an old kernel? I wish I knew the answer to that.

On the 5.4 kernel, I definitely saw exactly the same resullts from the old code to my version, since they do exactly the same thing, i.e. renice the kernel workqueue thread to -20. I did not test the case where we do not muck with the priority, but I will.

On newer kernels, the old method is already failing (since the workqueues are set up differently now), but I need to benchmark the differences to see if they even matter, since, as you say, they may no longer be needed. I will check that. But if we changed the code to just do nothing, we would have the same issue you raise with respect to older kernels.

mikechristie:
Oh yeah, sorry for the lack of updates. I looked up the priority inversion issue I mentioned. I think it came from or was similar to what's in drivers/md/md.c:md_thread:
        /*
         * md_thread is a 'system-thread', it's priority should be very
         * high. We avoid resource deadlocks individually in each
         * raid personality. (RAID5 does preallocation) We also use RR and
         * the very same RT priority as kswapd, thus we will never get
         * into a priority inversion deadlock.
         *
         * we definitely have to have equal or higher priority than
         * bdflush, otherwise bdflush will deadlock if there are too
         * many dirty RAID5 blocks.
         */

I don't remember the exact details for iscsi though. I think it's similar reasoning where the iscsi xmit thread needs to be able to always run since upper layers might be doing writes to free up memory.

gonzoleeman:
> Oh yeah, sorry for the lack of updates. I looked up the priority inversion issue I mentioned. I think it came from or was similar to what's in drivers/md/md.c:md_thread: /* * md_thread is a 'system-thread', it's priority should be very * high. We avoid resource deadlocks individually in each * raid personality. (RAID5 does preallocation) We also use RR and * the very same RT priority as kswapd, thus we will never get * into a priority inversion deadlock. * * we definitely have to have equal or higher priority than * bdflush, otherwise bdflush will deadlock if there are too * many dirty RAID5 blocks. */
> 
> I don't remember the exact details for iscsi though. I think it's similar reasoning where the iscsi xmit thread needs to be able to always run since upper layers might be doing writes to free up memory.

This makes sense for 5.* and older kernels. In my testing, I didn't see any deadlocks when I reset the priority to 0 (normal) on the workqueue thread, but I wasn't pushing my system very hard.

I agree with your assessment that we need to keep the priority setting as it is for older systems, which means leaving it in tact, for now.

gonzoleeman:
Ok, I've spent quite a bit of time analyzing performance testing, using fio, on various configurations and kernels.

One thing I've learned is that the new method (in some 5.* and all 6.* kernels) of setting workqueue priority using sysfs is not useful for purposes of iSCSI performance, for my tests. On my 6.8 kernel, I found that setting the priority via sysfs actually _hurt_ performance by about 10%!

But I am loathe to remove the current process-table scan, since older kernels may still need this, to avoid the priority deadlock @mikechristie  mentioned.

Just a reminder: the issue is that this current method of scanning the proc table just doesn't work on modern (6.8) kernels, as workqueues can now be passive, and not show up in the process table. In addition, it's slow, and it won't scale well to systems with lots of processes and lots of iSCSI connections.

My plan is to modify the iscsid priority-setting code so that if the "nice" value configured is zero, it will skipping setting it. I will also change the default to 0 and update iscsid.conf with a blurb about why it's zero, what that means, and that it will go away in general some time "soon".

This means that installations using new code will not do any priority tampering, since the default nice value will be zero. And existing node entries will continue to have a default of "-20", so they will continue to get the same behavior. And, in either case, storage administrators can modify the node or default "nice" value, if they wish.

I will prepare a new pull request, then close this one.

gonzoleeman:
Superseded by pull request #445

PR #444, opened 2024-03-16
iscsid: Rescan devices on relogin

mikechristie:
If a device goes to offline/transport-offline when we are disconnected the device's size can be set to zero if some other app forces it to get rescanned. Normally in the kernel when a device is set from offline to running again, it will rescan the device. However, because during a relogin the iscsi layer sets the state from offline/transport-offline via iscsi_unblock_session we don't hit that rescan. We actually don't want to hit the kernel's rescan right now, because it can block and that would block all iscsid operations.

This has us perform a rescan operation after a relogin from the scan process.

### DIFF
--- a/usr/initiator.c
+++ b/usr/initiator.c
@@ -1008,11 +1008,14 @@ void free_initiator(void)
 }
 
 static void session_scan_host(struct iscsi_session *session, int hostno,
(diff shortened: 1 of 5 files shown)


gonzoleeman (approved):
Looks good to me. Thank you.

PR #445, opened 2024-04-02
Modify how workqueue priority is set

gonzoleeman:
The open-iscsi code has been scanning the process table when it creates a new session (and hence a new host), in order to raise the priority of the workqueue transmit thread, to avoid a possible priority inversion with kernel threads like bdflush.

In modern kernels the workqueue transmit thread can be passive, meaning that it will not show up in the process table unless it is actively doing work. This means that we can no longer count on scanning the process table. But the good news is that those modern workqueue threads are designed differently and will not get deadlocked at their normal priority.

For now, we just optimize the code so that if the requested priority is zero, we do nothing, and we change the default value to zero, as well.

In the future the process-table-scanning code should be removed.

### DIFF
--- a/etc/iscsid.conf
+++ b/etc/iscsid.conf
@@ -214,8 +214,21 @@ node.session.queue_depth = 32
 # Note: For cxgb3i, you must set all sessions to the same value.
 # Otherwise the behavior is not defined.
 #
-# The default value is -20. The setting must be between -20 and 20.
-node.session.xmit_thread_priority = -20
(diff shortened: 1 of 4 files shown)


gonzoleeman:
This replaces pull request #441

gonzoleeman:
Having heard no objections or other comments, I'll merge this.

Note:t this changes the default "priority" (nice) value in iscsid.conf from "-20" to "0".

Issue #95, opened 2018-03-22
Race condition between access & mkdir


tasleson:
From the time it takes you to get the result of `access` to when you do the `mkdir`, the state of the FS may have changed.  Suggest alternative implementations which don't suffer from this.
``` C
int idbm_lock(void)
{
        int fd, i, ret;

        if (db->refs > 0) {
                db->refs++;
                return 0;
        }

        if (access(LOCK_DIR, F_OK) != 0) {
                if (mkdir(LOCK_DIR, 0660) != 0) {
                        log_error("Could not open %s: %s", LOCK_DIR,
                                  strerror(errno));
                        return ISCSI_ERR_IDBM;
                }
        }
...
```

gonzoleeman:
Please submit a pull request / patch (if not already done).

gonzoleeman:
Marking as resolved since a pull request has been merged.

uberjay:
The associated pull request did not fix this issue (it was merely mentioned in a PR comment that this issue was filed for tracking purposes). The access/mkdir race still exists!

gonzoleeman:
I'm not sure what you want here, but re-opening.

Please describe the scenario where an error can occur? Are you running two iscsid daemons? If not, then who is creating the lock directory between the time access() says it's not there and we do the mkdir()?

It seems like one solution to this not-really-a-problem would be to check access() again if mkdir fails with EEXIST. Do you have a better suggestion? I believe I asked for a pull request and never got one ...

gonzoleeman:
Note: there are at least half a dozen other places in open-iscsi where it does the access()/mkdir() sequence, not just for the lock directory.

uberjay:
The lock directory is in place to synchronize access by concurrent executions of `iscsiadm` -- if everything on disk were only read and written from iscsid, presumably there wouldn't be a need for a lock directory (as there would only be one process interacting with the on-disk state).

I don't know about any of the other access/mkdir sequenecs, but I presume they could be safe, as (maybe???) they are only done after acquiring the idbm lock? I'm not sure, as the only one I've seen occur is the lock directory.

To reproduce it, remove the lock directory and execute 2 concurrent iscsiadm commands which would need access to the on-disk iscsi state:

```
[root@localhost ~]# rm -rf /var/lock/iscsi
[root@localhost ~]# iscsiadm -m node & iscsiadm -m node
[1] 12629
iscsiadm: Could not open /var/lock/iscsi: File exists
[...]

PR #453, opened 2024-05-22
Better handle multiple iscsiadm commands

gonzoleeman:
The DB locking was racing, at times, since it did something like:

    if (access() != 0)
	    if (mkdir() != 0)
		    error

But when multiple processes are running, it is possible for a process to have access() fail, but then mkdir also fail, since another process has created the directory between the check and the attempt at creation. (Note that this error only seemed to occur on first use of iscsiadm, during testing.)

This change essentially makes EEXIST of the directory ok, with some other cleanups of the locking code (in two places).

### DIFF
--- a/libopeniscsiusr/idbm.c
+++ b/libopeniscsiusr/idbm.c
@@ -315,27 +315,41 @@ int _idbm_lock(struct iscsi_context *ctx)
 
 	db = ctx->db;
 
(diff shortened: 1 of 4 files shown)


gonzoleeman:
See issue #95

cleech:
OK, this all looks good to me :+1:

PR #454, opened 2024-05-29
systemd: Skip units when running in container

DaanDeMeyer:
iscsi might be installed in an image that can be booted both as a container and as a virtual machine (or bare metal). Let's condition out the units when running in a container as iscsi can't reliably do its job in a container anyways.

Without this commit, iscsi may fail to start when booting an image with iscsi installed in a systemd-nspawn container.


gonzoleeman:
Why did you withdraw this request? I haven't even looked at it yet.

BTW, there is code sitting around that would allow iscsid to run in a container because it would start respecting the network namespace it was in. But there was some resistance to its goal/results. Let me know if that interests you.

DaanDeMeyer:
I found another solution and don't have the interest to get this merged :(. Feel free to take these change and add them yourself if you'd like.

PR #457, opened 2024-06-24
Make it visible when memory allocation failure

ldzhong:
During the node discovery process, if there's a memory allocation failure occured in idbm_print(), an empty node DB file will be created silently.


gonzoleeman (approved):
Looks good to me.

PR #464, opened 2024-08-21
grammar nitpicks

mcpreston:


gonzoleeman:
I will consider these, but they are kind of silly. First of all, both "it's" and "its" are allowed, though "its" is more common these days. and the second commit fixes grammar that hardly matters IMHO.

gonzoleeman:
Thank you. I look forward to pull requests in the future that address more serious issues.

PR #465, opened 2024-08-23
Fix memory leak in iscsi_check_session_use_count

yanyan2015:
safe_logout is obtained by applying for strdup in cfg_get_string_param. 
When safe_logout is not a null pointer, there is no place to release it after use, which will cause the risk of memory leaks.


gonzoleeman:
Thank you.

Issue #475, opened 2024-09-19
clang build error:variable 'num_targets' set but not used


yuncang123:
Hi!When I use clang to build, I got this error:
```c
[   93s] discovery.c:698:6: error: variable 'num_targets' set but not used [-Werror,-Wunused-but-set-variable]
[   93s]   698 |         int num_targets = 0;
[   93s]       |             ^
[   93s] 1 warning and 1 error generated.
```

The variable 'num_targets' is only self added,and it's never used.Maybe it can be remove?

Issue #477, opened 2024-09-24
clang build error:field with variable sized type not at the end of a struct or class is a GNU extension


yuncang123:
```c
[   93s] In file included from io.c:36:
[   93s] In file included from ./initiator.h:31:
[   93s] ./mgmt_ipc.h:95:23: error: field 'stats' with variable sized type 'struct iscsi_stats' not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]
[   93s]    95 |                         struct iscsi_stats stats;
[   93s]       |                                            ^
[   93s] In file included from io.c:36:
[   93s] ./initiator.h:186:17: error: field 'rsp' with variable sized type 'iscsiadm_rsp_t' (aka 'struct iscsiadm_rsp') not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]
[   93s]   186 |         iscsiadm_rsp_t rsp;
[   93s]       |                        ^
[   93s] ./initiator.h:253:15: error: field 'reopen_qtask' with variable sized type 'queue_task_t' (aka 'struct queue_task') not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]
[   93s]   253 |         queue_task_t reopen_qtask;
[   93s]       |                      ^
```

yuncang123:
@gonzoleeman Hello!
I make a little case to test the problem,as follows:
iscsi_if.h

```c
#define ISCSI_STATS_CUSTOM_MAX		32
#define ISCSI_STATS_CUSTOM_DESC_MAX	64
#include <stdint.h>
struct iscsi_stats_custom {
	char desc[ISCSI_STATS_CUSTOM_DESC_MAX];
	uint64_t value;
};

struct iscsi_stats {
  uint32_t custom_length;
	struct iscsi_stats_custom custom[]
		__attribute__ ((aligned (sizeof(uint64_t))));
};
```

mgmt_ipc.h

```c
#include "iscsi_if.h"

typedef struct iscsiadm_rsp {
  union{
  struct ipc_msg_getstats {
      struct iscsi_stats stats;
			char custom[sizeof(struct iscsi_stats_custom) *
					ISCSI_STATS_CUSTOM_MAX];
		} getstats;
  }u;
}iscsiadm_rsp_t;
```

iscsiadm.c

```c
#include "mgmt_ipc.h"
#include<stdio.h>

int main(){
[...]

PR #478, opened 2024-09-24
fix 4 issues which are finded when building with clang 17

yuncang123:
Hi!When I build with clang 17,I got 4 errors with -Werror.
This pr fix issue #473 #474 #475 #476  .
You can take my modifying methods or use better methods,hopelly the bugs will be fixed.
Thanks!

### DIFF
--- a/usr/initiator.h
+++ b/usr/initiator.h
@@ -230,7 +230,7 @@ typedef struct queue_task {
 	iscsiadm_req_t req;
 	iscsiadm_rsp_t rsp;
 	int mgmt_ipc_fd;
-	int allocated : 1;
+	unsigned int allocated : 1;
 	/* Newer request types include a
 	 * variable-length payload */
 	void *payload;
--- a/usr/discovery.c
+++ b/usr/discovery.c
@@ -692,7 +692,6 @@ process_sendtargets_response(struct str_buffer *sendtargets,
 	char *end = text + str_data_length(sendtargets);
 	char *nul = end - 1;
 	char *record = NULL;
-	int num_targets = 0;
 
 	if (start == end) {
 		/* no SendTargets data */
@@ -741,7 +740,6 @@ process_sendtargets_response(struct str_buffer *sendtargets,
 					str_truncate_buffer(sendtargets, 0);
 					goto done;
 				}
-				num_targets++;
 			}
 			record = text;
 		}
@@ -763,7 +761,6 @@ process_sendtargets_response(struct str_buffer *sendtargets,
 				 record, record);
 			if (add_target_record(record + 11, text,
 					      drec, rec_list)) {
-				num_targets++;
 				record = NULL;
 				str_truncate_buffer(sendtargets, 0);
 			} else {
--- a/usr/discovery.c
+++ b/usr/discovery.c
@@ -32,6 +32,7 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 
+#include "iface.h"
 #include "local_strings.h"
 #include "types.h"
 #include "iscsi_proto.h"
@@ -46,7 +47,6 @@
 #include "transport.h"
 #include "iscsi_sysfs.h"
 #include "iscsi_ipc.h"
-#include "iface.h"
 #include "iscsi_timer.h"
 #include "iscsi_err.h"
 #if ISNS_SUPPORTED
--- a/usr/discoveryd.c
+++ b/usr/discoveryd.c
@@ -28,6 +28,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include "iface.h"
 #include "discovery.h"
 #include "idbm.h"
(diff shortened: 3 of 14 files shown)


gonzoleeman:
What procedure are you using to test building with clang? Can you be more specific? I'd like to try it myself.

usr/initiator.h
```diff
@@ -228,12 +228,14 @@ struct iscsi_ev_context {
 typedef struct queue_task {
 	iscsi_conn_t *conn;
 	iscsiadm_req_t req;
-	iscsiadm_rsp_t rsp;
```
gonzoleeman, line 231: I don't get this. The iscsiadm_rsp_t is not a variable-sized structure.

usr/initiator.h
```diff
@@ -228,12 +228,14 @@ struct iscsi_ev_context {
...
 	int mgmt_ipc_fd;
 	unsigned int allocated : 1;
 	/* Newer request types include a
 	 * variable-length payload */
 	void *payload;
+  /* field 'rsp' with variable sized type 'iscsiadm_rsp_t' 
+   * must be at the end of a struct or class is a GNU extension */
+  iscsiadm_rsp_t rsp;
```
gonzoleeman, line 238: This is indented incorrectly, as you can see from the diffs on github. If this field does indeed get moved, it needs to be indented correctly, and honestly I'd kind of like "req" to move with it, since they are paired somewhat (request, response).

usr/initiator.h
```diff
@@ -309,6 +304,15 @@ typedef struct iscsi_session {
...
 	queue_task_t *notify_qtask;
+ 
+	/* connection reopens during recovery */
+	int reopen_cnt;
+	int reopen_max;
+	iscsi_session_r_stage_e r_stage;
+	uint32_t replacement_timeout;
+  /* field 'reopen_qtask' with variable sized type 'queue_task_t'
+   * must be at the end of a struct or class is a GNU extension */
+  queue_task_t reopen_qtask;
```
gonzoleeman, line 315: Also indented incorrectly, and also based on iscsiadm_rsp_t being variable sized, which I don't see.

usr/mgmt_ipc.h
```diff
@@ -92,9 +92,11 @@ typedef struct iscsiadm_rsp {
 						ISCSI_STATS_CUSTOM_MAX)
 		struct ipc_msg_getstats {
 			struct iscsi_uevent ev;
-			struct iscsi_stats stats;
 			char custom[sizeof(struct iscsi_stats_custom) *
 					ISCSI_STATS_CUSTOM_MAX];
+      /* field 'stats' with variable sized type 'struct iscsi_stats' 
+       * must be at the end of a struct or class is a GNU extension */
+      struct iscsi_stats stats;
```
gonzoleeman, line 99: Also indented incorrectly. And why did this one get moved, when it doesn't seem to be part of the structures you claim are variable-sized?

gonzoleeman (changes requested):
the first four commits look fine, but the last one, attempting to handle structures with variable sizes, is not ok.

Can you remove that one and resubmit (force push would work, or a new branch)? I'd be glad to approve those first 4 commits, even though I can't build with clang, since the all make sense. Then we can discuss my issues with the last commit separately.

P.S. I'd still like to know how to test with clang myself. I tried setting CC=clang, but my build failed almost immediately, making libopeniscsiusr.

yuncang123:
> the first four commits look fine, but the last one, attempting to handle structures with variable sizes, is not ok.
> 
> Can you remove that one and resubmit (force push would work, or a new branch)? I'd be glad to approve those first 4 commits, even though I can't build with clang, since the all make sense. Then we can discuss my issues with the last commit separately.
> 
> P.S. I'd still like to know how to test with clang myself. I tried setting CC=clang, but my build failed almost immediately, making libopeniscsiusr.

Really thanks for your review!I've already use force push to remove the fifth commit and resubmit.

My answers to your questions:
Q1:How do I test building with clang ?
A1:I am actually participating in an open source project activity organized by the openEuler operating system community. The purpose of this project is to make the software support both GCC and LLVM builds. I used the online build platform recommended by the community: Open Build Service (OBS). What I need to do is to branch the project (which have already used Project Config to switch the compiler toolchain to clang) and solve the errors presented by the build.Here's the address:https://build.tarsier-infra.isrc.ac.cn/package/show/Mega-LLVM:24.03/open-iscsi .You can just click the red 'failed' to see the build log.

Q2:The reason why 'setting CC=clang, but build failed almost immediately, making libopeniscsiusr' ？
A2:I'm sorry that I don't know the exact reason for this, but I guess you can refer to this blog (translating the page into English may help you understand) to switch the compiler toolchain to clang if you use rpmbuild.

Thanks again for correcting my mistake！
Next, we should probably discuss issues separately under #477 .

gonzoleeman:
Thank you for your help!

Issue #484, opened 2024-10-17
Login hangs with node.session.scan = manual


sbogomolov:
I have noticed that iSCSI Kubernetes Persistent Volumes stopped working after upgrading `open-iscsi` from `2.1.9` to `2.1.10`. I was also able to reproduce the issue on my desktop using the same version of `open-iscsi`, so it is not distribution-specific. When `kubelet` is trying to attach a volume it does this:

```
iscsiadm -m iface -I default -o show
iscsiadm -m discoverydb -t sendtargets -p 192.168.70.20:3260 -I default -o new
iscsiadm -m discoverydb -t sendtargets -p 192.168.70.20:3260 -I default --discover
iscsiadm -m node -p 192.168.70.20:3260 -T iqn.2000-01.com.synology:nas.authentik-redis.2edf3217d57 -I default -o update -n node.session.scan -v manual
iscsiadm -m node -p 192.168.70.20:3260 -T iqn.2000-01.com.synology:nas.authentik-redis.2edf3217d57 -I default --login
```

With `open-iscsi` `2.1.10` the last command hangs and never completes.

- If I set `node.session.scan = auto` - login works.
- If I downgrade `open-iscsi` to `2.1.9` - login works with both `node.session.scan = manual` and `node.session.scan = auto`.

Context and initial investigation: https://github.com/siderolabs/extensions/issues/490

gonzoleeman:
I believe this behavior changed with the following commit:

> 48e683c858fa ('iscsid: Rescan devices on relogin (#444)')

by Mike Christie on 4/2/2024. It fixed an issue with resizing and having iscsid block.

Perhaps you can just enable node.session.scan=auto (which is the default)?

gonzoleeman:
@mikechristie do you wish to comment?

sbogomolov:
> Perhaps you can just enable node.session.scan=auto (which is the default)?

I don't think I can do that. This behavior is controlled by k8s. `kubelet` explicitly sets `node.session.scan=manual` for the specific target right before logging in.

gonzoleeman:
On Oct 18, 2024, at 10:30 AM, Sergii Bogomolov ***@***.***> wrote:
> 
> 
> Perhaps you can just enable node.session.scan=auto (which is the default)?
> 
> I don't think I can do that. This behavior is controlled by k8s. kubelet explicitly sets node.session.scan=manual for the specific target right before logging in.
> 

That seems like the bug. Why do they do that?

[...]

PR #485, opened 2024-10-18
iscsid: Fix hang during login with scan=manual

mikechristie:
This fixes a regression added in:

48e683c ('iscsid: Rescan devices on relogin (#444)

 For scan=manual we were not sending a response to callers like iscsiadm so they would hang.

See issue:

https://github.com/open-iscsi/open-iscsi/issues/484


sbogomolov:
I can confirm that this fixes the issue.

gonzoleeman:
Thank you, as always, Mike.

PR #506, opened 2025-04-08
Makefile: fix "No rule to make target 'iscsiuio/Makefile.in" issue

MaXinjian:
| cd iscsiuio; autoreconf --install
| make: *** No rule to make target 'iscsiuio/Makefile.in', needed by 'iscsiuio/Makefile'.  Stop. | make: *** Waiting for unfinished jobs....
| libtoolize: putting auxiliary files in '.'

iscsiuio/Makefile.in will automatically generated when generate iscsiuio/configure by "cd iscsiuio; autoreconf --install", no need to explicitly specify

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ user: deprecation_msg iscsiuio/Makefile
 	@echo
 	@echo "Read README file for detailed information."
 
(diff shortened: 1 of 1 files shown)


gonzoleeman:
I will go ahead an merge this, but using Makefile's is no longer supported, so if it works, or doesn't, it's on you. Trying using ninja/meson.

PR #511, opened 2025-09-03
iscsi: delay reconnect until interface has valid IP.

shmsingh:
When an interface goes down and later comes back up, there is a window where the link is up but no IP address has been assigned yet. In this state, an immediate reconnect attempt can lead to unexpected behavior.

Consider this scenario:
On the initiator side with two interfaces (re0 and re1) connected to the target via a switch, consider the following sequence:
- re1 goes down and its IP address is moved to re0 as part of failover.
- re0 communicates with the target, which then caches re0's IP along with its hardware address in arp table.
- Later, re1 comes back online. When the link first comes up, there is a short window where no IP is assigned to re1.
- If iSCSI attempts to reconnect during this window, it will (by design) fall back to using another interface with a valid IP. In this case, the packet is sent with re1's hardware address but re0's IP address.
- The target updates its cache and ends up with two different hardware addresses mapped to the same IP address.
- This inconsistent cache entry prevents the session from being restored even after re1 eventually receives its IP.

Additionally, due to this ARP mismatch, ULP's such as RDS that monitor ARP change to detect connection state will also be impacted.

This patch adds a check to ensure that the interface has a valid IPv4 or IPv6 address before attempting to reconnect. If no usable address is found, the reconnect is delayed and retried later.

### DIFF
--- a/usr/io.c
+++ b/usr/io.c
@@ -29,6 +29,8 @@
 #include <netinet/tcp.h>
 #include <arpa/inet.h>
 #include <sys/uio.h>
+#include <ifaddrs.h>
+#include <netinet/in.h>
 
 #include "iface.h"
 #include "types.h"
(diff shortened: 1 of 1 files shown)


gonzoleeman (approved):

gonzoleeman:
Looks good to me. Thank you for your contribution!

shmsingh:
Thanks @gonzoleeman  for helping with the review and accepting my patch.

PR #523, opened 2026-01-24
Ensure that iscsi.service is enabled when iscsid.service is enabled, and move Before= to a more appropriate target

gordonmessmer:
iscsi.service exists to provide a mechanism to send an asynchronous signal to iscsid, so that it can reload its configuration on changes to the network status. Because it is an asynchronous signal, specifying that the unit must finish Before any other unit is ineffective.

Removing this setting will allow systems that do not mount any filesystems from iscsi targets to boot without unnecessary delays.


gordonmessmer:
Would you consider this smaller change? (which I am also sending to the Fedora/RHEL maintainer)

PR #527, opened 2026-05-25
iscsi: validate interface IP against target address family

pkannoju:
Currently, iscsi_conn_iface_has_ip() treats the presence of either an IPv4 or IPv6 address on the interface as sufficient for interface readiness.

In environments where IPv6 is temporarily enabled during interface bring-up, the kernel may briefly generate an IPv6 link-local address (fe80::/10). This can cause the readiness check to succeed even when the iSCSI target portal is IPv4-based and the interface does not yet have a valid IPv4 address assigned.

As a result, the initiator may proceed with TCP connection setup while binding traffic to an interface that does not yet have a usable IPv4 address, leading to MAC/IP mismatch behavior during source address selection.

Fix this by validating the interface address against the target portal address family. For IPv4 targets, only IPv4 addresses are considered valid. Likewise, for IPv6 targets, only IPv6 addresses are considered valid.

This prevents transient IPv6 link-local addresses from incorrectly satisfying interface readiness checks for IPv4 iSCSI connections.

### DIFF
--- a/usr/io.c
+++ b/usr/io.c
@@ -207,10 +207,10 @@ static int bind_src_by_address(int sockfd, char *address)
  * return 1: if the interface has a valid IP.
  */
 static bool
-iscsi_conn_iface_has_ip(char *iface_name)
+iscsi_conn_iface_has_ip(char *iface_name, int expected_family)
 {
 	struct ifaddrs *ifaddr = NULL, *ifa;
-	int family, status = 0;
+	int status = 0;
 
 	/*
 	 * if getifaddrs() fails, something went wrong, requeue the request.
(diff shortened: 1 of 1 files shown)


mikechristie:
> the kernel may briefly generate an IPv6 link-local address (fe80::/10).

If the user is using IPv6 addrs, will the kernel create this tmp address and could you the issue still?

pkannoju:
Yes, the kernel can still generate an IPv6 link-local address (`fe80::/10`) even when the user intends to use IPv6 connectivity. In such cases, the behavior is expected because IPv6 link-local addresses are part of normal IPv6 interface initialization.

The intent of the proposed patch is not to block or reject IPv6 link-local addresses. Instead, the goal is to ensure that the interface readiness validation matches the address family of the target portal.

For example:

* If the iSCSI target portal is IPv4-based, the interface should have a valid IPv4 address before the connection is attempted.
* If the iSCSI target portal is IPv6-based, the interface should have a valid IPv6 address before the connection is attempted.

So for an IPv6 target:

* the socket family is `AF_INET6`
* the connect path uses `tcp_v6_connect()`
* and the interface validation also checks for IPv6 addresses

In that scenario, the proposed change does not alter the expected IPv6 behavior. It simply ensures that the address family used for interface validation remains consistent with the address family used for the actual iSCSI connection.

mikechristie:
Ok. Patch looks ok to me.

pkannoju:
Thank you very much @mikechristie .

gonzoleeman:
Looks good. Merged. Thank you! And thank you to Mike for reviewing.

PR #528, opened 2026-05-29
avoid possible double free of found in idbm_rec_update_param

Cav4ever:
The found buffer allocated for TYPE_INT_LIST handling is freed in the updated path, but the pointer is not cleared afterwards.

Later, check_password_param() may redirect control flow back to setup_passwd_len via goto, causing idbm_rec_update_param() to re-enter the update path with found still containing the stale freed pointer value.

In that case, the second pass may reach the updated label without reinitializing found, leading to free(found) being executed again on the already freed pointer.

The issue can be avoided by clearing found after freeing it.

### DIFF
--- a/usr/idbm.c
+++ b/usr/idbm.c
@@ -1167,8 +1167,10 @@ free_tmp:
 
 updated:
 	strlcpy((char*)info[i].value, value, VALUE_MAXVAL);
-	if (found)
+	if (found) {
 		free(found);
+		found = NULL;
+	}
 
 #define check_password_param(_param) \
 	if (!passwd_done && !strcmp(#_param, name)) { \


mikechristie:
The patch looks ok to me.

Did you hit this or were you manually reviewing code or using a AI or static review tool?

Cav4ever:
> The patch looks ok to me.
> 
> Did you hit this or were you manually reviewing code or using a AI or static review tool?

I found it while using Xmaze AI, an AI code security agent developed by Xmirror.

gonzoleeman:
I will merge this because it seems correct. But I fear this project may need to add some requirements about AI's involvement in bugs and fixes, as the kernel has started to do.

I think that, at the very least, folks should include how they found the bug they are suggesting a fix for, whatever form that takes. And fixes generated by AI should be annotated as such, as well, IMHO.

All that said, thank you for your fix.

PR #531, opened 2026-06-17
Fix iscsi_conn_iface_has_ip error handling

mikechristie:
This fixes 2 issues with:

Commit 7e5b2757fed8 ("iscsi: delay reconnect until interface has valid IP. (#511)")

1. If iscsi_conn_iface_has_ip fails we need to close the socket.
2. This also fixes a tiny coding style issue where we want over 80 cols.

### DIFF
--- a/usr/io.c
+++ b/usr/io.c
@@ -255,7 +255,8 @@ iscsi_conn_iface_has_ip(char *iface_name, int expected_family)
 	return status;
 }
 
-static int bind_conn_to_iface(iscsi_conn_t *conn, struct iface_rec *iface, int expected_family)
+static int bind_conn_to_iface(iscsi_conn_t *conn, struct iface_rec *iface,
+			      int expected_family)
 {
 	struct iscsi_session *session = conn->session;
 
@@ -326,17 +327,18 @@ iscsi_io_tcp_connect(iscsi_conn_t *conn, int non_blocking)
 		return -1;
 	}
 
-	if (bind_conn_to_iface(conn, &conn->session->nrec.iface, ss->ss_family))
-		return -1;
+	if (bind_conn_to_iface(conn, &conn->session->nrec.iface,
+			       ss->ss_family)) {
+		rc = -1;
+		goto close_sock;
+	}
 
 	onearg = 1;
 	rc = setsockopt(conn->socket_fd, IPPROTO_TCP, TCP_NODELAY, &onearg,
 			sizeof (onearg));
 	if (rc < 0) {
 		log_error("cannot set TCP_NODELAY option on socket");
-		close(conn->socket_fd);
-		conn->socket_fd = -1;
-		return rc;
+		goto close_sock;
 	}
 
 	/* optionally set the window sizes */
@@ -396,6 +398,11 @@ iscsi_io_tcp_connect(iscsi_conn_t *conn, int non_blocking)
 	if (non_blocking)
 		set_non_blocking(conn->socket_fd);
 	rc = connect(conn->socket_fd, (struct sockaddr *) ss, sizeof (*ss));
+	return rc;
+
+close_sock:
+	close(conn->socket_fd);
+	conn->socket_fd = -1;
 	return rc;
 }
 


mikechristie:
Adding @gulams because he's good at reviewing patches and he will probably have to add it to OL :) Gulam, if it's not you can you tag the person that will handle it?

gulams (commented):
Looks good to me.
Reviewed-by: Gulam Mohamed <gulam.mohamed@oracle.com>

usr/io.c
```diff
@@ -255,7 +255,8 @@ iscsi_conn_iface_has_ip(char *iface_name, int expected_family)
 	return status;
 }
 
-static int bind_conn_to_iface(iscsi_conn_t *conn, struct iface_rec *iface, int expected_family)
+static int bind_conn_to_iface(iscsi_conn_t *conn, struct iface_rec *iface,
+			      int expected_family)
```
gonzoleeman, line 259: For the record, I do not currently have any rules about not going over 80 columns. It's been a long time since we had real terminals with only 80 columns, and I like the readability (and sometimes searchability) that having the code on one line can add.

But this is with a bigger commit that fixes issues, so it's fine with me.

usr/io.c
```diff
@@ -326,17 +327,18 @@ iscsi_io_tcp_connect(iscsi_conn_t *conn, int non_blocking)
 		return -1;
 	}
 
-	if (bind_conn_to_iface(conn, &conn->session->nrec.iface, ss->ss_family))
-		return -1;
+	if (bind_conn_to_iface(conn, &conn->session->nrec.iface,
+			       ss->ss_family)) {
```
gonzoleeman, line 331: I would actually try to match to rest of the function, i.e.:

```
rc = bind_conn_to_iface( <stuff> );
if (rc < 0) {
    log_error( <it might be nice to have an error message here?>);
    goto close_sock;
}
```
I do believe bind_conn_to_iface() returns 0 or -1.

usr/io.c
```diff
@@ -397,6 +399,11 @@ iscsi_io_tcp_connect(iscsi_conn_t *conn, int non_blocking)
 		set_non_blocking(conn->socket_fd);
 	rc = connect(conn->socket_fd, (struct sockaddr *) ss, sizeof (*ss));
 	return rc;
+
```
gonzoleeman, line 402: This might be a dumb question, but can the above "connect()" fail? If so, shouldn't it also close the socket connection and set the fd to -1?

gonzoleeman (commented):
I'd like your feedback on my comments.

And thank you for your submission!

usr/io.c
mikechristie: Agree. @gulams You were going to push this downstream right? Could you also handle upstream here as well?

usr/io.c
mikechristie: Yes and yes.

gonzoleeman:
Are updates coming?

gulams:
> Are updates coming?

@gonzoleeman, Yes, I will send it shortly

gonzoleeman:
We will soon be tagging a new release. Any chance you can get this fix in before that @gulams ?

gulams:
> We will soon be tagging a new release. Any chance you can get this fix in before that @gulams ?

@gonzoleeman, I will try get this done over the weekend. Sorry for the late.

gonzoleeman:
> > We will soon be tagging a new release. Any chance you can get this fix in before that @gulams ?
> 
> @gonzoleeman, I will try get this done over the weekend. Sorry for the late.

No worries. Thanks.

usr/io.c
gulams: > I would actually try to match to rest of the function, i.e.:
> 
> ```
> rc = bind_conn_to_iface( <stuff> );
> if (rc < 0) {
>     log_error( <it might be nice to have an error message here?>);
>     goto close_sock;
> }
> ```
> 
> I do believe bind_conn_to_iface() returns 0 or -1.

@gonzoleeman , I think this error message is not required as the callers of this function are already logging the error message. Let me know what do your comments.

usr/io.c
gulams: > This might be a dumb question, but can the above "connect()" fail? If so, shouldn't it also close the socket connection and set the fd to -1?

@gonzoleeman , I verified the code couple of times. Yes, the connect can fail but this failure is taken care and the fd is closed by the callers. Here is the detail:

1. In ep_connect() there are 3 cases:
   A. iscsi_conn_connect(), 2 cases:
     a. __session_conn_reopen():Not required as we do queue_delayed_reopen() where ultimately it will disconnect
        and do a reconnect
     b. __session_login_task(): Here also we do queue_delayed_reopen(). So, same as above case 'a'
   B. iscsi_create_leading_conn(), 1 case: Already taken care, so not needed
2. For direct connect() call, this is the call flow where its already taken care:
	iscsi_create_leading_conn() --> iscsi_io_connect() --> iscsi_io_tcp_connect() --> connect()
	[close(fd) is done here in iscsi_create_leading_conn]

@gonzoleeman , So, I think both the log error message and the close(fd) are not required. I think you can go ahead and merge. If you think I am missing anything, can you please let me know?

usr/io.c
gonzoleeman: I believe you are saying that bind_conn_to_iface() prints an error message when it returns failure, and a quick check shows that to be the case.

So you can skip this suggested change.

gulams: > bind_conn_to_iface

Actually I was saying about this following print in "iscsi_io_connect()" function :

} else if (rc < 0) {
		log_error("cannot make connection to %s: %s",  <<<<<<<<<<<
			  conn->host, strerror(errno));
		close(conn->socket_fd);
		ret = 0;
		goto done;
	} else if (log_level > 0) {

When bind_conn_to_iface() returns -1, the function iscsi_io_tcp_connect() also returns -1 to iscsi_io_connect() and this will print the above log_error().
And as you said, bind_conn_to_iface() also prints a log_error message whenever it returns -1.

usr/io.c
gonzoleeman: looks ok

PR #532, opened 2026-06-22
Use iscsi-gen-initiatorname in iscsi-init

Mr-Bossman:


etc/systemd/iscsi-init.service.template
```diff
@@ -16,4 +16,4 @@ WantedBy=systemd-remount-fs.service
 [Service]
 Type=oneshot
 RemainAfterExit=no
-ExecStart=/usr/bin/sh -c 'echo "InitiatorName=`@SBINDIR@/iscsi-iname`" > /etc/iscsi/initiatorname.iscsi'
+ExecStart=/usr/bin/sh -c 'iscsi-gen-initiatorname'
```
Mr-Bossman: It might also be good to make the use of `iscsi-gen-initiatorname` consistant and use it for `make install_iname`, but having the script find  `iscsi-iname` when not in `/sbin` was not trivial.

gonzoleeman: Sorry, but no.

I agree using iscsi-gen-initiatorname is a good idea. Our distribution has:

> ExecStart=@BINDIR@/iscsi-gen-initiatorname

But you can't use "sh -c" because you are relying on PATH, and that's security hole. No telling what PATH may be set to.

The meson script to convert certain special pathnames in template files into real names is already present, so removing it here doesn't save any complexity. These pathnames are filled in at build time, when we know what pathname we are using, i.e. we don't have to "find" anything on a live system.

As far as the "make" targets, they are deprecated in favor of meson. They are only still around, to help some transition, and for a limited time. If you are using them, please transition to meson. I'll be glad to help you get started if you would like.

So please close or fix this pull request.

utils/iscsi-gen-initiatorname.sh.template
```diff
@@ -16,8 +16,7 @@ INAME_DIR="@HOMEDIR@"
 INAME_FILE="$INAME_DIR/initiatorname.iscsi"
 
 # our default IQN prefix
-DEFAULT_IQN_PREFIX="iqn.1996-04.de.suse:01"
-
+DEFAULT_IQN_PREFIX="@IQN_PREFIX@"
 #
```
gonzoleeman, line 21: I like this idea.

utils/Makefile
```diff
@@ -18,7 +18,7 @@ HOMEDIR ?= $(etcdir)/iscsi
 
 RULESDIR ?= $(etcdir)/udev/rules.d
 
-IQN_PREFIX ?= "iqn.2016-04.com.open-iscsi"
```
gonzoleeman, line 21: I do not care about Makefiles and will not test them. If you resubmit this I will likely merge your changes here, with that huge caveat.

gonzoleeman (commented):
this one might be okay, with stated caveat

etc/systemd/iscsi-init.service.template
gonzoleeman: I'm sorry, ours has "@SBINDIR@/iscsi-gen-initiatorname". Apologies for visual cut-n-paste error.

Mr-Bossman: > But you can't use "sh -c" because you are relying on PATH, and that's security hole. No telling what PATH may be set to.

Interesting didn't realize that. good to know

utils/Makefile
Mr-Bossman: Oh ok i was wondering why there is both meson and make files. the makefiles do work for the utils directory still so ill keep this change

cleech:
Thanks for working on this @Mr-Bossman  
@gonzoleeman, I asked Jesse to look at making this script work for us at RH. Thanks for reviewing.  
I think this is looking good.

utils/iscsi-gen-initiatorname.sh.template
```diff
@@ -16,7 +16,7 @@ INAME_DIR="@HOMEDIR@"
 INAME_FILE="$INAME_DIR/initiatorname.iscsi"
 
 # our default IQN prefix
-DEFAULT_IQN_PREFIX="iqn.1996-04.de.suse:01"
+DEFAULT_IQN_PREFIX="@IQN_PREFIX@"
```
cleech, line 19: Can you also update doc/iscsi-gen-initiatorname.8.template to use @IQN_PREFIX@?  Thank you.

```
--- a/doc/iscsi-gen-initiatorname.8.template
+++ b/doc/iscsi-gen-initiatorname.8.template
@@ -36,7 +36,7 @@ Force overwrite of existing initiator name, if present.
 .TP
 .BI [-p] \fIIQN-PREFIX\fP
 Use \fIIQN-PREFIX\fP as the prefix to the IQN generated,
-instead of the default of \fBiqn.1996-04.de.suse:01\fP.
+instead of the default of \fB@IQN_PREFIX@\fP.
 .SH FILES
 .TP
 @HOMEDIR@/initiatorname.iscsi
```

cleech (approved):

PR #534, opened 2026-06-23
iscsiuio: Fix -Waddress-of-packed-member

Mr-Bossman:
Fix -Waddress-of-packed-member warnings in
uip-neighbor.c:uip_neighbor_out by using memcpy().

Fix warings in uip-arp.c:uip_arp_update by changing the type of ipaddr to const u8_t * instead of u16_t *, and using memcmp() for comparison.

### DIFF
--- a/iscsiuio/src/uip/uip_arp.c
+++ b/iscsiuio/src/uip/uip_arp.c
@@ -119,7 +119,7 @@ void uip_arp_timer(void)
 }
 
 /*----------------------------------------------------------------------------*/
-static void uip_arp_update(u16_t *ipaddr, struct uip_eth_addr *ethaddr)
+static void uip_arp_update(const u8_t *ipaddr, struct uip_eth_addr *ethaddr)
 {
 	u8_t i;
 	struct arp_entry *tabptr;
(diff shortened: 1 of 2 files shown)


iscsiuio/src/uip/uip_arp.c
```diff
@@ -136,8 +136,8 @@ static void uip_arp_update(u16_t *ipaddr, struct uip_eth_addr *ethaddr)
 
 			/* Check if the source IP address of the incoming packet
 			   matches the IP address in this ARP table entry. */
-			if (ipaddr[0] == tabptr->ipaddr[0] &&
-			    ipaddr[1] == tabptr->ipaddr[1]) {
+			if (memcmp(ipaddr, tabptr->ipaddr,
```
gonzoleeman, line 139: I'm not crazy about using memcmp() to compare numbers. But this is low-level code that knows how the numbers are laid out. But I guess I can live with it, since it will remove those annoying compiler warnings.

gonzoleeman (approved):
Looks good to me.

iscsiuio/src/uip/uip_arp.c
Mr-Bossman: we could alternatively use `memcpy` like in `uip_neighbor_out`, but because `ipaddr` is packed in this case an array and `tabptr->ipaddr` is also an array, the resulting code should be identical in functionality.

Mr-Bossman:
@gonzoleeman are there any updates on this?

gonzoleeman:
Merged. Thank you.

PR #535, opened 2026-06-23
Fix security issues recently discovered by Keith at Linneman Labs

gonzoleeman:
Fix two security-related issues in open-iscsi.


keithlinneman:
Just leaving a quick note here - I sent you an email.

usr/discovery.c
```diff
@@ -126,6 +126,35 @@ int discovery_isns_set_servername(char *address, int port)
 	return 0;
 }
 
+/*
+ * iqn_name_valid -- is the supplied IQN name valid?
+ *
```
mikechristie: If we are doing docbook style then I think you don't want the newline.

Also I noticed I messed up and I think we are supposed to do 2 stars "/**", but in many places I did 1 "/*" like you did above.

usr/discovery.c
```diff
@@ -612,6 +653,12 @@ add_target_record(char *name, char *end, discovery_rec_t *drec,
 		log_error("TargetName %s too long, ignoring", name);
 		return 0;
 	}
+
+	if (!iqn_name_valid(name)) {
+		log_error("Discovery TargetName invalid: \"%s\"", name);
+		return 0;
+	}
+
```
mikechristie, line 632: We can also add record manually:

iscsiadm -m node -o new -T target_name ....

so I think you want add checks in exec_node_op(). Or put it in iscsiadm:main() when we get the target name passed in and are about to use it in the call

main()
{
....
                rec = idbm_create_rec(targetname, tpgt, ip, port, iface, 1);

And then we can also create records via the fw discovery method. Checkout exec_fw_disc_op. I think you want a check in there or exec_disc_op_on_recs.

I think you also want to check iscsistart.c:main(). When we create a session by just passing in the info, then later we can use that info to create records.

Also I noticed above for iSNS you check the initiator name. Do you want to check the initiator name in other places to be consistent even though it will not cause the issue described? Or can you hit the issue with initiator names as well. The bug description mentioned target name only so I was not sure.

iscsiuio/src/unix/iscsid_ipc.c
```diff
@@ -1120,7 +1120,7 @@ static void *iscsid_loop(void *arg)
 			continue;
 		}
 
-		if (!mgmt_peeruser(iscsid_opts.fd, user) || strncmp(user, "root", PEERUSER_MAX)) {
+		if (!mgmt_peeruser(s2, user) || strncmp(user, "root", PEERUSER_MAX)) {
 			close(s2);
 			ILOG_ERR(PFX "Access error: non-administrative connection rejected");
 			break;
```
mikechristie: The s2 part of the fix looks good to me.

Do we want to continue here though? If we break then a non root user could do a request and that would fail correctly now, but then iscsiuio would exit. We wouldn't be able to handle future requests or if we are mid multi request operation we wouldn't be able to handle it.

usr/discovery.c
gonzoleeman: I will fix this

iscsiuio/src/unix/iscsid_ipc.c
gonzoleeman: Yes. Keith brought up the same point.

usr/discovery.c
gonzoleeman: Thank you Mike for your thorough review. It's important to get this correct, and I appreciate your help.

I removed the check for initiator name. That was just an error on my part. I don't see that as a valid attack vector. If the attacker can change /etc/iscsi/initiatorname.iscsi, then they've already broken into root-owned files and could do anything anyway. I added a check for target name there.

I added checks in iscsiadm.c and iscsistart.c.

I will submit my update soon.

gonzoleeman:
I have updated my branch and force-pushed it. Here's a list of the changes:

Changes in V2:
============

* in general
 * do dot log invalid IQNs (except in debug mode at high
   debug level)
 * better wording in IQN validaty errors, to make it clear
   what is being done
 * moved iqn_name_valid from discovery.c to initiator_common.c,
   no longer static via initiator.h

* in iscsid_ipc.c:
 * change "break" to "continue" in the case of a validation error
   (this was never hit before because validation always succeeded)

* in discovery.c
 * stop printing invalid IQN name (except in debug)
 * fix printing of "bad" character (if any) in IQN name
 * change debug printing of bad name from level 7 to level 8 (for debug)
 * fix docbook for new function
 * in discovery_isns_query(), no need to check out initiator name,
   change to checking target name

* in iscsiadm.c
 * in main: check passed in targetname before calling idbm_create_rec
 * in exec_disc_op_on_recs(), check returned target name (for "update" or
   "new")

* in iscsistart.c
 * in main, check passed in target name

iscsiuio/src/unix/iscsid_ipc.c
gonzoleeman: I believe I fixed this.

gonzoleeman:
@mikechristie @cleech @keithlinneman : any more feedback, or can I go ahead?

usr/discovery.c
gonzoleeman: fixed with new pull request.

keithlinneman:
Looks good from my side. All look resolved cleanly.

Issue #540, opened 2026-08-09
Breaking Change in Patch Version 2.1.11 -> 2.1.12


FrostbittenKing:
Hi, after debugging a broken k3s node on my arch linux server, I tracked this down to a rename of a config parameter in iscsid.conf:
`node.session.sess_reopen_log_freq`  which was renamed from `node.session.conn_reopen_log_freq`.
This led to obscure very hard to debug errors in kubernetes, which prevented scheduling / attaching of volumes in longhorn via iscsi.
It took me hours to find the source in the logs for the issue. Imho, this was a very bad decision to do in a patch level change, which took my sunday afternoon.
Why wasn't this at least in the patch notes? At least mentioning it would have probably led to a latest news  post on archlinux.org that requires manual intervention.
I'm sorry I don't understand this decision to do this in a patch level change from 2.1.11 -> 2.1.12, this is not very professional. :(.
I would appreciate patch notes updates, so others are informed about this, and not having to deal with this debugging an issue after a kubernetes node update for hours.

gonzoleeman:
I'm sorry that you had an issue you had to spend time debugging.

I missed the fact that changing the name of this configuration variable would break existing installations. The commit that created conn_reopen_log_freq was only two weeks and 3 commits before it was changed to sess_reopen_log_freq. In retrospect, I should have asked that the old value was also accepted for some grace period, perhaps with a warning about deprecation.

I will be glad to retag the current code with a higher version number, indication such a change happened, and add a note the the Changes file. Or I could add a change that would accept the old value for a while, as mentioned above. But both of these approaches will not change the fact that 2.1.11 and 2.1.12 are out there, and if you use this configuration variable you will have an issue when upgrading to 2.1.12.

Also, it seems harsh that the daemon will not start up or run if an unknown configuration variable is set ...

gonzoleeman:
I am currently considering one way of mitigating this issue, and would appreciate some feedback.

Right now, if libopeniscsiusr /idbm.c detects an unknown attribute in the Node DB (e.g. "/var/lib/iscsi/nodes/iqn*/..."), it considers it an error and fails to process the database entry (in_idbm_rec_update_param()) :

```
759 unknown_value: 
 760                                 _error(ctx, "Got unknown data type %d "
 761                                        "for name '%s', value '%s'",
 762                                        recs[i].data, recs[i].name,
 763                                        recs[i].value);
 764                                 rc = LIBISCSI_ERR_BUG;
[...]

PR #542, opened 2026-08-17
Fix reopen log freq change

gonzoleeman:
A recent change to  the node.session attribute from conn_reopen_log_freq to sess_reopen_log_freq made it clear that treating unknown attributes as a fatal error was too harsh. So change that to a warning.

Also, in this case, treat the old attribute name as a deprecated alias for the new name.

I don't believe these is any need to print a warning when doing this aliasing. Having the old attribute name present doesn't hurt anything, and the first time they update their DB entry or rediscover a node, the old value will be replaced with the new one.

Tested with node DB entries from 2.1.11 and from 2.1.12


cleech:
I've tested this out, it works like I expected and I think getting this merged and in a new tagged release is the best we can do for #540

Issue #541, opened 2026-08-14
use of uninitialized `if_hwaddr` in `find_vlan_dev`


dirkmueller:
https://github.com/open-iscsi/open-iscsi/commit/996269f9ea214e776eb81f9ff538da788c873a14#diff-6734bcb2d7b22d4b4368c8b5a6a4fea187c67e34af754bcaa4bb9178e6e6532bL202

removed the only remaining initialization of the stack local variable `if_hwaddr`, which is otherwise completely uninitialized. 

Right afterwards, `ioctl(sockfd, SIOCGIFHWADDR, &if_hwaddr)` is executed on this uninitialized struct, which most likely contains complete garbage in `ifr_name`. this error isn't caught because the return value of the ioctl() call is not checked at all, but instead silently assumed to be passing, and afterwards the struct members are accessed, like for example

`if (if_hwaddr.ifr_hwaddr.sa_family != ARPHRD_ETHER)` 

which accesses uninitialized memory and causes undefined behavior. 

It is not entirely clear to me what the function tries to do, but either a lot of the code needs to be removed (anything that handles if_hwaddr) or the previous strlcpy removal needs to be reworked somehow. the linux kernel "tools/virtio/vhost_net_test.c" as well as other open source code that I can find are always memset'ing the entire struct to zero before invoking this ioctl() and also check the return code before continuing. 

the way this was introduced in 2.1.12 means it is now I think broken.

cleech:
While that commit left an uninitialized struct ifreq, I think it was the previous commit in this file 72c2b97 "IPv6 support for iBFT iSCSI boot" that first probably broke vlan support and passed a NULL physdev to find_vlan_dev.  I'll take a look at this, but I don't have a ready test for vlan booting.

gonzoleeman:
Closing this, since the fix was just merged.

PR #545, opened 2026-08-19
iscsi_net_util: fix broken VLAN support in find_vlan_dev

cleech:
Commit 72c2b97c introduced a bug by passing physdev to find_vlan_dev() before it was initialized (physdev was set after the call instead of before). This was then made worse by commit 996269f which removed the netdev parameter entirely, leaving if_hwaddr.ifr_name uninitialized.

This fix:
- Restores the netdev parameter to find_vlan_dev()
- Re-adds the strlcpy() to initialize if_hwaddr.ifr_name
- Fixes the call order to set physdev before calling find_vlan_dev() in both net_setup_netdev_ipv4() and net_setup_netdev_ipv6()

I've briefly tested this with iscsistart -N with an iBFT table containing a VLAN
- 2.1.11 segfaults
- 2.1.12 fails to find the VLAN device
- 2.1.12 + this fix correctly configures the VLAN device

### DIFF
--- a/usr/iscsi_net_util.c
+++ b/usr/iscsi_net_util.c
@@ -185,7 +185,7 @@ free_ifni:
 	return 0;
 }
 
-static char *find_vlan_dev(int vlan_id) {
+static char *find_vlan_dev(char *netdev, int vlan_id) {
 	struct ifreq if_hwaddr;
 	struct ifreq vlan_hwaddr;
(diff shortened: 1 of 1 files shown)


cleech:
this should fix #541

gonzoleeman:
Excellent. Thank you.

PR #549, opened 2026-09-01
Make shell scripts POSIX compatible and add a shellcheck workflow

ncopa:
Make all scripts work with POSIX shell so we don't need bash.

Add a github workflow to check scripts with shellcheck.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,20 @@
+name: ShellCheck
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install ShellCheck
+        run: sudo apt-get update && sudo apt-get install --yes shellcheck
+      - name: Check shell scripts
+        run: |
+          git -c color.ui=false grep -lE '^#!.*(ba)?sh([[:space:]]|$)' | \
+            xargs -r shellcheck
--- a/etc/initd/initd.debian
+++ b/etc/initd/initd.debian
@@ -1,4 +1,5 @@
 #! /bin/sh
+# shellcheck disable=SC2034
 ### BEGIN INIT INFO
 # Provides:
 # Required-Start:
@@ -15,6 +16,7 @@ PIDFILE=/run/iscsid.pid
 
 [ -x "$DAEMON" ] || exit 0
 
+# shellcheck disable=SC1091
 . /lib/lsb/init-functions
 
 if [ ! -d /sys/class/ ]; then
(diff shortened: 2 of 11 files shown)


gonzoleeman:
It may take me a while to evaluate this request. I am not sure I'm comfortable changing "bash" to "sh". It works fine for my distribution, since "sh" is just a symlink to "bash". But I worry other distributions could have an issue with this change. Is the reason you want to do this to support musl? Does your platform not have bash?

I also do not know anything about spellcheck, nor about GitHub workflows. It looks like you've set it up for ubuntu only, using apt-get. I'm not sure how I feel about adding a workflow I don't use and can't easily run. Perhaps that should be a separate pull request, and include other distributions? FTR, I downloaded shellcheck on Tumbleweed, version 0.11.0, and  several of the scripts still fail shellcheck after your changes, mostly about putting double quotes around variable usage.

ncopa:
> Is the reason you want to do this to support musl?

No. Reason is that in Alpine Linux we use busybox ash as default shell (it is around 60-70kb compared to bash 1MB+). We have for a long time had downstream [patches](https://gitlab.alpinelinux.org/alpine/aports/-/blob/3.24-stable/main/open-iscsi/0001-make-shell-scripts-POSIX-compatible.patch?ref_type=heads) to remove the dependency of bash. I now took the time to clean those up a bit.

> Does your platform not have bash?

Not by default, but you can always `apk add bash` on Alpine Linux, but those scripts are simple and does not really depend on any bash specific features. The cleanups are trivial. We simply want remove an unnecessary dependency.

> I also do not know anything about spellcheck

`shellcheck`(https://www.shellcheck.net/) is a standard shell linter to find common (potential) shell script problems. It can be used to detect bash specific features and ensure that script is actually POSIX compatible.

> nor about GitHub workflows. It looks like you've set it up for ubuntu only, using apt-get. I'm not sure how I feel about adding a workflow I don't use and can't easily run.

The idea is that the github workflow runs when contributors opens pull requests. If the changes passes the `shellcheck`  linter it will turn green, and if it fails, it will be red. I added this mostly to prove that  I don't introduce any syntax problems and to prevent future changes to unintentionally re-introduce bash specific changes.

> ... include other distributions?

I don't think that makes any sense. We only want verify that the shell scripts passes a linter, which distro that happens on does not matter. I may make sense to run a test build on different distros though, but that is out of scope for this PR.

> FTR, I downloaded shellcheck on Tumbleweed, version 0.11.0, and several of the scripts still fail shellcheck after your changes, mostly about putting double quotes around variable usage.

Ok. I suppose we can do it in two steps:
- use posix shell
- fix shellcheck

ncopa:
BTW...

> But I worry other distributions could have an issue with this change

That is why I used shellcheck.

ncopa:
Superseded by https://github.com/open-iscsi/open-iscsi/pull/552 which should be easier to review.

PR #550, opened 2026-09-01
gitignore: ignore generated build artifacts

ncopa:


gonzoleeman:
If you use mason to build, and build in a subdirectory, as is standard practice, many of these .gitignore patterns would not be needed. But I see no harm in adding them.

PR #551, opened 2026-09-01
iscsiadm: handle getopt errors directly

ncopa:
getopt_long signals parse failures through its return value, while optopt may be zero for an invalid long option on musl.

Handle '?' and ':' immediately so invalid options and missing arguments are consistently rejected.

ref: https://gitlab.alpinelinux.org/alpine/aports/-/work_items/4802


ncopa:
this should also fix downstream: https://github.com/siderolabs/extensions/blob/main/storage/iscsi-tools/patches/musl-fixes.patch

gonzoleeman:
I tested out this change, and it does seem to clean up the error messages from getopt. Thank you.

PR #553, opened 2026-09-03
network code ioctl handling improvements

### LINKED ISSUE #547: fix the remaining part of #541
The original ticket pointed out a missing memset() on the struct and missing error return check on ioctl(). neither one was fixed in the merged PR, so the part of the report on `if()` still accessing potentially uninitialized fields (like hw_addr) still holds.

see #541

cleech:
Replace explicit memset() calls with struct initialization at declaration, replace magic number 1 with ARPHRD_ETHER constant, and add missing error checks for ioctl(SIOCGIFHWADDR) calls in find_vlan_dev function.

fixes #547


gonzoleeman (approved):
Looks good to me

PR #557, opened 2026-09-21
iscsid: honor REPORT LUNS DATA HAS CHANGED again

brianatpurestorage:
Commit 48e683c ("iscsid: Rescan devices on relogin (#444)") made the per-device rescan and the host-wide `- - -` scan mutually exclusive — the host scan sits in the `else` branch of the new `rescan` flag. That was aimed at the relogin path, but the AEN handler shares `session_scan_host()` and was switched to `rescan=true` as well, which silently disabled REPORT LUNS DATA HAS CHANGED (ASC/ASCQ 0x3f/0x0e).

A target sending it now gets only a capacity refresh of the LUNs the initiator already has. `iscsi_sysfs_rescan_device()` writes to `/sys/class/scsi_device/H:0:T:L/device/rescan`, which exists only for LUNs that already have a device node, so a newly mapped LUN is never found — only the host scan issues a REPORT LUNS.

The worst case is a target mapping its *first* LUN to an already connected initiator: with no LUNs mapped there are no `H:0:T:L` entries at all, `iscsi_sysfs_for_each_device()` iterates an empty set, and the AEN produces no sysfs writes whatsoever. The session never drops, so nothing else recovers it and the LUN stays invisible until someone runs `iscsiadm -m session -R` by hand. We hit this with array-side ACLs that allow a connection before any volume is mapped.

Affects 2.1.10 through 2.1.13 — 48e683c landed 2024-04-02 and was first released in 2.1.10 on 2024-05-31.

This makes the three operations independent flags instead of an either/or, so the AEN path can refresh existing devices and scan for new ones. Every other caller keeps its current behavior.

It also passes the sid to `session_scan_host()` explicitly. `iscsi_async_session_creation()` calls it with a NULL session for sessions the HBA firmware created, and 48e683c added a `session->id` dereference below the early return that used to cover that case, so iscsid would crash on those transports whenever `node.session.scan` was auto.

A follow-up patch to scan for new LUNs when the I_T nexus is re-established will be sent separately.


mikechristie:
Hey, Thanks for the patch and sorry of the regression.

I realized I made the mistake a couple months ago and actually made a similar patch that had flags to allow control what is done. In the end I thought it was not needed and just did:

https://github.com/mikechristie/open-iscsi/commit/54e766470448305db32b0ff7eeb738e6148059fa 

I'm fine with your approach. I had originally been looking into something similar because I was trying to handle the issue when iscsid exits while trying to do the initial iscsi login (when iscsid restarts we don't scan) which might be a special case of what you are looking into (https://github.com/mikechristie/open-iscsi/commits/iscsid-validate).

Are you adding the extra infrastructure for the follow up patch?

brianatpurestorage:
Thanks — and no need to apologize. It took us a while to work out why our AEN had stopped doing anything, but once we found 48e683c858fa it was obvious.

> Are you adding the extra infrastructure for the follow up patch?

Yes, that's exactly it. Your one-liner fixes the regression on its own, and it's the more precisely scoped fix for it — strictly, 0x3f/0x0e is about the LUN inventory rather than capacity, so the rescan-existing half isn't really needed to honor the AEN.

We went the flags route because the second patch is the functionality we're actually after, and it needs all three operations to be independent. On relogin after an I_T nexus loss we want to bring the existing devices back online, rescan them, *and* scan for LUNs that appeared while the path was down. The `bool` can't express that combination, which is the same reason the AEN path broke: the two branches are mutually exclusive.

So it's your call which way round to do it, and either is fine by us:

1. Take your one-liner here as the regression fix, and we bring the flags with the follow-up where they carry their own weight.
2. Keep this as-is and the follow-up becomes a small patch on top.

Two things ride along on the signature change that we'd want to keep somewhere either way:

- `iscsi_async_session_creation()` calls `session_scan_host()` with a NULL session, and 48e683c858fa added a `session->id` dereference below the early return that used to cover that case. iscsid segfaults on CAP_FW_DB transports when node.session.scan is auto. Passing the sid explicitly fixes it. Happy to split that into its own patch — it's unrelated to the AEN and probably wants to be standalone regardless.
- A behavior difference worth flagging: with node.session.scan=manual our version still rescans existing devices on the AEN, where yours and the pre-48e683c code do nothing at all. We can drop that if you'd rather manual mean manual.

> the issue when iscsid exits while trying to do the initial iscsi login (when iscsid restarts we don't scan)

We think the follow-up covers that. On restart, `sync_session()` → `MGMT_IPC_SESSION_SYNC` → `iscsi_sync_session()` sets `R_STAGE_SESSION_REOPEN` and calls `session_conn_reopen()`, so the relogin lands in the same `setup_full_feature_phase()` branch the follow-up teaches to scan. Less a special case than the same missing scan reached by a different route.

On testing: we've run this patch as posted against our array and confirmed it resolves the problem. Mapping the first volume to an already connected initiator now brings the LUN up off the AEN, where before the message was parsed and then produced nothing at all. That was with node.session.scan at the default auto.

Two gaps to be straight about. The manual-mode difference noted above is untested — easy for us to run, we just haven't yet. The CAP_FW_DB NULL deref we can't exercise at all, as we don't have an offload HBA here, so that one is code inspection only.

